PR #3, opened 2017-03-24
Missing config filename causes logstash to not start.

KALRONG:
In the docker-compose.yml the name of the config file for the -f options passed to logstash is missing causing it to not find the file and therefore crashing with an error.


jlruizmlg:
hi @KALRONG by default logstash only apply the *.conf in the configuration, if you allocate something different than config files in this folder it's a bad practice, please verify that.

PR #4, opened 2017-03-24
Add postfix on wazuh container build, add postfix start to run.sh and…

KALRONG:
… update docker-compose to allow map postfix configuration folder.


jlruizmlg:
Hi @KALRONG thanks for your contribution, like always!!!

PR #7, opened 2017-04-24
Configurable URL for kibana plugin

elisiano:
Added WAZUH_KIBANA_PLUGIN_URL variable to wait-for-it.sh so that kibana plugin installation URL can be overwritten (the default still points to https://packages.wazuh.com/wazuhapp/wazuhapp-2.0_5.3.0.zip).
This is useful in (most likely corporate) environments where there is no egress to the internet or users just prefer to mirror all software locally.

Could also be useful to test a different version of the plugin.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -60,6 +60,8 @@ services:
     depends_on:
       - elasticsearch
     entrypoint: sh wait-for-it.sh elasticsearch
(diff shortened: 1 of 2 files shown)


jlruizmlg:
Hi @elisiano great idea, thanks for your contribution!.

PR #12, opened 2017-08-02
fix docs

davidkarlsen:
Fix docs to match current state


jlruizmlg:
HI @davidkarlsen that is correct, thanks!.

PR #14, opened 2017-08-17
Fix ES hostname resolution for kibana/logstash

peteralcock:
Without linking the containers with explicitly declared container name mappings, the "elasticsearch" hostname is not being resolved by the kibana or logstash containers. This fixes that.


jlruizmlg:
Hi @peteralcock thanks for your contribution!

PR #21, opened 2017-09-24
Adding Nginx container

mjflve:
Hi,

This PR will add the ability to use Nginx as a reverse proxy of Kibana and provides SSL + authentication, please review.

Thanks,


jlruizmlg:
Thanks @mjflve LGTM! looks good to me, let me test!

PR #24, opened 2017-10-22
Automatic password generation for wazuh API

davinerd:
I think hardcoded credentials are bad, especially if users don't think about it and expose this docker container outside.

I added a small fix to auto generate the password. Username is hardcoded (well, at least it's not easily guessable!)

### DIFF
--- a/wazuh/config/init.bash
+++ b/wazuh/config/init.bash
@@ -5,6 +5,12 @@
 #
 source /data_dirs.env
 
+API_USER="wazadmin"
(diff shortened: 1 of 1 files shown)


jlruizmlg:
Hi @davinerd sorry for the delay in our response, in the last commit we include the connection with the API automatically

- https://github.com/wazuh/wazuh-docker/blob/master/kibana/config/wait-for-it.sh#L33

If we apply your pull request we will lose this new feature.

Issue #40, opened 2018-02-27
Add filebeat as volume


jlruizmlg:
In order to do changes in ```filebeat.yml``` easily we can add filebeat folder as mount folder and let ```docker-compose.yml``` mount the folder through the volumes volumes section.

RANGERBEE:
what is correct syntax for multiple logstash hosts?  the load balancing switch is confusing me.

I have this as my filebeat.yml currently - not sure it is correct:

```yml
filebeat:
 prospectors:
  - input_type: log
    paths:
     - "/var/ossec/data/logs/alerts/alerts.json"
    document_type: wazuh-alerts
    json.message_key: log
    json.keys_under_root: true
    json.overwrite_keys: true

output:
 logstash:
   # The Logstash hosts
   hosts: ["logstash:5000", "10.0.1.194:5514"]
#   ssl:
#     certificate_authorities: ["/etc/filebeat/logstash.crt"]

```

RANGERBEE:
I want to duplicate all events into multiple logstash indexers

RANGERBEE:
found this

[https://github.com/elastic/beats/issues/1035](url)

jlruizmlg:
Hi @RANGERBEE you have the merged the code to map the folder /etc/filebeat as an external folder and customize your own filebeat.yml.

About your question, let us test it and return to you as soon as possible.

RANGERBEE:
This is the reason I wanted to run a filebeat official container and monitor a host mounted volume to the alerts .logs on ossec server.  Filebeat currently will not ship to more than one Logstash , except for load balancing or "failover" configuration.

RANGERBEE:
FYI: I had to output to Kafka in order to get to my second Logstash instance - due to the limitations by design in Logstash and Beats.

SitoRBJ:
Hello @RANGERBEE,

We are sorry for the delay.  Did you manage to send the events to a second Logstash without load balancing? If so, if you want you can inform us of the process to add it to the repository or even create your own PR.

If not, we will work on it as soon as possible. 

Again we regret the delay and thank you for your patience. 

Best regards,

Alfonso Ruiz-Bravo

jesuslinares:
Hi @RANGERBEE ,

Please, don't hesitate to re-open the issue if you still have the problem, we'll be pleased to help you!.

Thanks.

PR #41, opened 2018-02-27
Issue 40.1

SitoRBJ:
Fix issue #40 . 

With the intention of making changes in a simpler way in filebeat.yml, we give the option to mount it as a volume.


jlruizmlg:
LGTM, thanks @SitoRBJ

PR #42, opened 2018-03-09
Add a mount point for custom Wazuh configuration files

JPLachance:
Having a way to mount custom configuration files in the Wazuh manager container will be really useful!

Why do we need a mount point for those files?
* I first tried to mount the `ossec.conf` file at the `/var/ossec/etc/ossec.conf` path, but on the first boot of the container, the [`if [ ! -e "${DATA_PATH}/${ossecdir}" ]`](https://github.com/wazuh/wazuh-docker/blob/master/wazuh/config/run.sh#L38) condition becomes false for the `etc` folder and the initial Wazuh setup in the container fails (the `etc` folder contains only the `ossec.conf` file and some SSL keys).
* Then, I tried to mount the `ossec.conf` file at the `/var/ossec/etc-template/ossec.conf` path, it worked on the initial install, but since I use a persistent storage for the `/var/ossec/data` folder (an AWS EBS drive mounted in my container), when I try to update the configuration and try to reboot my container to mount the new configuration file, the  [`if [ ! -e "${DATA_PATH}/${ossecdir}" ]`](https://github.com/wazuh/wazuh-docker/blob/master/wazuh/config/run.sh#L38) condition is false for the `etc` folder, so my new file in `etc-template` doesn't get copied in `etc`.

So finally, by creating a mount point for configuration files, mounting the files works with a newly created Wazuh container or with an existing one that is rebooting.

I already tested these changes by overriding the `run.sh` file in my container, it works :)

Changes:
\+ Add a mount point for custom Wazuh configuration files

\+ Add documentation for that mount point

\! Fix the currently supported Wazuh and Elasticsearch versions in the README

\! Fix the ossec_shutdown function in the `wazuh/config/run.sh` file (/var/ossec/data/bin is not valid)

### DIFF
--- a/wazuh/config/run.sh
+++ b/wazuh/config/run.sh
@@ -12,8 +12,13 @@
 #
 
 source /data_dirs.env
+
 FIRST_TIME_INSTALLATION=false
-DATA_PATH=/var/ossec/data
+
+WAZUH_INSTALL_PATH=/var/ossec
+DATA_PATH=${WAZUH_INSTALL_PATH}/data
+
+WAZUH_CONFIG_MOUNT=/wazuh-config-mount
 
 print() {
     echo -e $1
@@ -29,6 +34,9 @@ exec_cmd() {
     eval $1 > /dev/null 2>&1 || error_and_exit "$1"
 }
 
+exec_cmd_stdout() {
+    eval $1 2>&1 || error_and_exit "$1"
+}
 
 edit_configuration() { # $1 -> setting,  $2 -> value
     sed -i "s/^config.$1\s=.*/config.$1 = \"$2\";/g" "${DATA_PATH}/api/configuration/config.js" || error_and_exit "sed (editing configuration)"
@@ -75,11 +83,28 @@ then
   fi
 fi
 
-#Enabling ossec-authd.
+##############################################################################
+# Copy all files from $WAZUH_CONFIG_MOUNT to $DATA_PATH and respect
+# destination files permissions
+#
+# For example, to mount the file /var/ossec/data/etc/ossec.conf, mount it at
+# $WAZUH_CONFIG_MOUNT/etc/ossec.conf in your container and this code will
+# replace the ossec.conf file in /var/ossec/data/etc with yours.
+##############################################################################
+if [ -e "$WAZUH_CONFIG_MOUNT" ]
+then
+  print "Identified Wazuh configuration files to mount..."
+
+  exec_cmd_stdout "cp --verbose -r $WAZUH_CONFIG_MOUNT/* $DATA_PATH"
+else
+  print "No Wazuh configuration files to mount..."
+fi
+
+# Enabling ossec-authd.
 exec_cmd "/var/ossec/bin/ossec-control enable auth"
 
 function ossec_shutdown(){
-  ${DATA_PATH}/bin/ossec-control stop;
+  ${WAZUH_INSTALL_PATH}/bin/ossec-control stop;
 }
 
 # Trap exit signals and do a proper shutdown
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ In addition, a docker-compose file is provided to launch the containers mentione
 
 ## Current release
 
-Containers are currently tested on Wazuh version 3.1.0 and Elastic Stack version 6.1.0. We will do our best to keep this repository updated to latest versions of both Wazuh and Elastic Stack.
+Containers are currently tested on Wazuh version 3.2.0 and Elastic Stack version 6.2.1. We will do our best to keep this repository updated to latest versions of both Wazuh and Elastic Stack.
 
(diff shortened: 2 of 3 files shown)


mjflve:
Hi @JPLachance,

First of all thanks for your contribution. Can you please tell us some use cases for this in particular?. If you set the `/var/ossec/data` volume from the beginning you can acquire similar behavior, it would put several files like the ossec.conf, rules, and decoders on the host machine mapped folder with the ability to edit those files:

```
  wazuh:
    image: wazuh/wazuh
    hostname: wazuh-manager
    restart: always
    ports:
      - "1514:1514/udp"
      - "1515:1515"
      - "514:514/udp"
      - "55000:55000"
    networks:
        - docker_elk
    volumes:
      - /var/ossec:/var/ossec/data:Z
#      - my-path:/etc/postfix:Z
#      - my-path:/etc/filebeat
    depends_on:
      - elasticsearch
```

Please let us know. Thanks!

Best Regards,

JPLachance:
Hi @mjflve,

Like I said in the description, I first tried to mount the `ossec.conf` file at the `/var/ossec/etc/ossec.conf` path, which is the same as trying to mount the file at the `/var/ossec/data/etc/ossec.conf` path, because `/var/ossec/etc` is a symlink for `/var/ossec/data/etc`.

That being said, mounting the `ossec.conf` file at the `/var/ossec/data/etc/ossec.conf` path is not working because of the initial install loop in the `run.sh` file.

In the [run.sh](https://github.com/wazuh/wazuh-docker/blob/master/wazuh/config/run.sh#L38) file, you have this:
```bash
for ossecdir in "${DATA_DIRS[@]}"; do
  if [ ! -e "${DATA_PATH}/${ossecdir}" ]
  then
    print "Installing ${ossecdir}"
    exec_cmd "mkdir -p $(dirname ${DATA_PATH}/${ossecdir})"
    exec_cmd "cp -pr /var/ossec/${ossecdir}-template ${DATA_PATH}/${ossecdir}"
    FIRST_TIME_INSTALLATION=true
  fi
done
```

The `DATA_DIRS` array contains the `etc` folder. So, you will end up doing that:
```bash
  if [ ! -e "/var/ossec/data/etc" ]
  then
    print "Installing etc"
    exec_cmd "mkdir -p $(dirname /var/ossec/data/etc)"
    exec_cmd "cp -pr /var/ossec/etc-template /var/ossec/data/etc"
    FIRST_TIME_INSTALLATION=true
  fi
```

So, why is mouting a file `/var/ossec/data/etc` before the initial install breaking the `etc` folder? Simply because the ` exec_cmd "cp -pr /var/ossec/etc-template /var/ossec/data/etc"` line won't be executed because the ` if [ ! -e "/var/ossec/data/etc" ]` condition will be false.

Then, mounting a file in `/var/ossec/data/etc` before that loop is preventing the loop from properly copying files from `/var/ossec/etc-template` into `/var/ossec/data/etc` which results in a broken `/var/ossec/data/etc` folder (almost all the files that should be there are not).

Also, another problem with mounting the file directly into `/var/ossec/data/etc` is the fact the permissions of the files you will mount won't be the ones expected by Wazuh/Ossec. For example, the `ossec.conf` file belongs to `root:ossec`, but if you try to mount it with:
```yaml
  wazuh:
    image: wazuh/wazuh
    hostname: wazuh-manager
    restart: always
    ports:
      - "1514:1514/udp"
      - "1515:1515"
      - "514:514/udp"
      - "55000:55000"
    networks:
        - docker_elk
    volumes:
      - my-local-conf/ossec.conf:/var/ossec/data/etc/ossec.conf:Z
    depends_on:
      - elasticsearch
```
Then the `ossec.conf` file will belong to `root:root` (I didn't test with Docker compose, but in Kubernetes when mounting a ConfigMap in a Pod, the file would belong to `root:root`). This is not problematic with the `ossec.conf` file, but it can be with other files if they belong to a user other than root.

With my solution, you can mount any file you want in `/wazuh-config-mount/*` and they will be copied in `/var/ossec/data/*` without their permissions, so the files at the destination will keep their permissions settings.

I hope this explains the why and the how 😄 

I didn't know how to name the `/wazuh-config-mount`, so if you have a better idea, I'm opened to suggestions!

wazuh/config/run.sh
```diff
@@ -29,6 +34,9 @@ exec_cmd() {
     eval $1 > /dev/null 2>&1 || error_and_exit "$1"
 }
 
+exec_cmd_stdout() {
```
JPLachance: I added that because it's very useful to see which file was copied where by the `exec_cmd_stdout "cp --verbose -r $WAZUH_CONFIG_MOUNT/* $DATA_PATH"` command!

mjflve:
Hi @JPLachance,

Thanks for the clarification and your contribution.

Best Regards,

PR #44, opened 2018-04-13
Update Dockerfile

augustine-urolime:
nginx copy command fix


mjflve:
Hi @augustine-urolime, 

Good catch, thanks for your contribution, we appreciate it!

Best Regards,

PR #47, opened 2018-07-21
Fixed bad permissions in filebeat.yml

rafadvega:


jlruizmlg:
Hi @rafadvega you are right! LGTM thanks for your contribution!!

PR #48, opened 2018-07-26
Fix wazuh/wazuh-kibana image version to 3.4.0_6.3.1

wirabdillah:
`wazuh-kibana` image version version is wrong. It should be `3.4.0_6.3.1`


jlruizmlg:
@wirabdillah thanks! good catch!

Issue #57, opened 2018-10-02
Allow custom commands or scripts to be run before services are started


forana:
`/entrypoint.sh` both configures and starts the services, meaning that, in order to run `ossec-control` actions and have the changes be picked up, there are two hacky solutions I see:

1. Either run those via `docker exec` and then either restart the container or the services within (neither of which are great for an automated installation), or
2. Volume over a modified copy of [wazuh.runit.service](https://github.com/wazuh/wazuh-docker/blob/master/wazuh/config/wazuh-api.runit.service) and add custom commands before `service wazuh-manager start`, which means that future changes to that file will not be picked up.

There should be a way to do this explicitly.

SitoRBJ:
Hello @forana,

We are sorry for the delay. 

This issue is solved in your PR #58

Thank you for your contribution, we proceed to close the issue. 

Best regards,

Alfonso Ruiz-Bravo

PR #58, opened 2018-10-02
Allow custom scripts or commands before service start

forana:
Addresses #57

This method is arguably only slightly less hacky than the two workarounds I added in #57, but it's a sounder strategy for automated installation than either of them.

### DIFF
--- a/wazuh/config/entrypoint.sh
+++ b/wazuh/config/entrypoint.sh
@@ -125,7 +125,5 @@ do
   echo "Executing command \`${CUSTOM_COMMAND}\`"
   exec_cmd_stdout "${CUSTOM_COMMAND}"
 done
-  exec_cmd_stdout "${CUSTOM_COMMAND}"
-done
 
 /sbin/my_init 
--- a/wazuh/config/entrypoint.sh
+++ b/wazuh/config/entrypoint.sh
@@ -113,4 +113,16 @@ trap "ossec_shutdown; exit" SIGINT SIGTERM
 
 chmod -R g+rw ${DATA_PATH}
 
+##############################################################################
+# Interpret any passed arguments (via docker command to this entrypoint) as
+# paths or commands, and execute them.
+#
+# This can be useful for actions that need to be run before the services are
+# started, such as "/var/ossec/bin/ossec-control enable agentless".
+##############################################################################
+for CUSTOM_COMMAND in "$@"
+do
+  exec_cmd_stdout "${CUSTOM_COMMAND}"
+done
+
 /sbin/my_init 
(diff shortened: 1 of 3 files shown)


wazuh/config/entrypoint.sh
```diff
@@ -113,4 +113,16 @@ trap "ossec_shutdown; exit" SIGINT SIGTERM
...
 
+##############################################################################
+# Interpret any passed arguments (via docker command to this entrypoint) as
+# paths or commands, and execute them.
+#
+# This can be useful for actions that need to be run before the services are
+# started, such as "/var/ossec/bin/ossec-control enable agentless".
+##############################################################################
+for CUSTOM_COMMAND in "$@"
+do
```
migruiz4: Hi @forana
I have tested it and works perfectly. Great job.
The only modification I think this needs, is printing some extra information about what commands are being executed, to make it more visible in the docker logs, for example adding this code.

```suggestion
do
  echo "Executing command \`${CUSTOM_COMMAND}\`"
  exec_cmd_stdout "${CUSTOM_COMMAND}"
done
```
Would make the output look like this:
```
wazuh_1 | Executing command `echo hello world`:
wazuh_1 | hello world
wazuh_1 | Executing command: `ls /var/ossec/`:
wazuh_1 | api
wazuh_1 | api-template
wazuh_1 | backup
wazuh_1 | bin
wazuh_1 | data
wazuh_1 | etc
wazuh_1 | etc-template
wazuh_1 | framework
wazuh_1 | integrations
wazuh_1 | lib
wazuh_1 | logs
wazuh_1 | logs-template
wazuh_1 | queue
wazuh_1 | queue-template
wazuh_1 | ruleset
wazuh_1 | ruleset-template
wazuh_1 | stats
wazuh_1 | stats-template
wazuh_1 | tmp
wazuh_1 | var
wazuh_1 | wodles
```
This solution can't print characters like ' or " without scaping then with \\, but this way we can know which command the output corresponds to.

Thank you very much for developing this new feature.
We appreciate all the contributions that help us improve our code.

Best regards,
Miguel

forana: That makes sense to me. I had it in there originally and removed it after some internal waffling - I've applied your suggestion to add it back.

migruiz4:
Hi @forana,
Thank you for applying the changes so quickly, but while reviewing again the entrypoint.sh, I realized now it looks like this:
```
for CUSTOM_COMMAND in "$@"
do
  echo "Executing command \`${CUSTOM_COMMAND}\`"
  exec_cmd_stdout "${CUSTOM_COMMAND}"
done
  exec_cmd_stdout "${CUSTOM_COMMAND}"
done
```
Can you please fix that little issue?

Regards,
Miguel

forana:
argh, sorry about that. should be fixed now @migruiz4

migruiz4:
Looks good to me!
Thank you for your contribution @forana.

PR #73, opened 2018-11-26
Using etc folder as a mounting point

havidarou:
Hello,

this PR uses `etc` folder as a mounting point for Wazuh persistent data option.

It fixes an issue where, after registering agents, `ossec-authd` replaces the soft link with a `client.keys`, thus rendering the persistent `client.keys` copy useless.

Regards.


jesuslinares:
Hi,

We should review what will happen with the _internal_options.conf_ file: If we upgrade the manager, it will use the file inside the mounting point. So, if there a missing setting, it will fail.

PR #79, opened 2018-12-18
Updating entrypoing.sh so upgrading is posible

### LINKED ISSUE #76: Review upgrade
The upgrade is failing due to:

**.templates.db files**
It will be fixed en Wazuh v3.8.0 (https://github.com/wazuh/wazuh/issues/2050) but we need a workaround:
The following files must be removed before the upgrade:
- /var/ossec/queue/db/.template.db
- /var/ossec/var/db/.template.db

**internal_options.conf**
Since we save the internal_options.conf in etc, it will not be overwritten during an upgrade. If Wazuh expects new settings in that file, it will fail.
We need to remove it and push the new one.

LuisGi93:
It fixes  #76. The correct internal_options is copied over the outdated. It also removes /var/ossec/queue/db/.template.db so upgrade is posible.

### DIFF
--- a/wazuh/config/entrypoint.sh
+++ b/wazuh/config/entrypoint.sh
@@ -53,6 +53,14 @@ for ossecdir in "${DATA_DIRS[@]}"; do
   fi
 done
 
+
+if [  -e ${WAZUH_INSTALL_PATH}/etc-template  ]
+then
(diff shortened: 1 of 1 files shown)


SitoRBJ:
Hello team,

This PR  is duplicated (#80).

We proceed to close this PR because the other is done on the last numbered branch and is preferable. 

Great job.

Regards,

Alfonso Ruiz-Bravo

PR #106, opened 2019-01-30
Add Wazuh-Elasticsearch

SitoRBJ:
Hello team,

This PR solves #108 and creates the Wazuh-Elasticsearch module. This is done to properly separate the tasks concerning each container. The configurations of each container only depend on itself and not on another container. 

To this end, we have transferred the modifications made to the Elasticsearch container from the Kibana container.

 We have also added the option of configuring API credentials using environment variables so that they are not embedded in the code. 

We have performed tests of various types:

- Clean deployment.
- Deployment with new API credentials.
- Deployment enabling volumes for Elasticsearch and Wazuh.
- Registration of agents. 
- Verification of the persistence of agent registration, configuration changes and alerts through the use of volumes.
- Verification of the functioning of the environment after removing the Elasticsearch container and creating it again. 

All tests have been successful. 

Best regards,

Alfonso Ruiz-Bravo

### DIFF
--- a/elasticsearch/config/entrypoint.sh
+++ b/elasticsearch/config/entrypoint.sh
@@ -22,7 +22,7 @@ if [ "$1" = 'elasticsearch' -a "$(id -u)" = '0' ]; then
 	set -- su-exec eulasticsearch "$@"
 	ES_JAVA_OPTS="-Des.network.host=$NETWORK_HOST -Des.logger.level=$LOG_LEVEL -Xms$HEAP_SIZE -Xmx$HEAP_SIZE"  $@ &
 else
-	$r@ &
+	$@ &
 fi
 
 ./load_settings.sh &
--- a/elasticsearch/config/entrypoint.sh
+++ b/elasticsearch/config/entrypoint.sh
@@ -1,30 +1,28 @@
 #!/bin/bash
+# Wazuh App Copyright (C) 2019 Wazuh Inc. (License GPLv2)
 
-set -m
+# For more information https://github.com/elastic/elasticsearch-docker/blob/6.5.4/build/elasticsearch/bin/docker-entrypoint.sh
 
-# Add elasticsearch as command if needed
-if [ "${1:0:1}" = '-' ]; then
-	set -- elasticsearch "$@"
-fi
+set -e
 
-if [ "$NODE_NAME" = "" ]; then
-	export NODE_NAME=$HOSTNAME
-fi
+# Files created by Elasticsearch should always be group writable too
+umask 0002
 
-if [ "x${ELASTICSEARCH_URL}" = "x" ]; then
-  el_url="https://elasticsearch:9200"
-else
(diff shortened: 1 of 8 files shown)


elasticsearch/config/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+	export NODE_NAME=$HOSTNAME
+fi
+
+if [ "x${ELASTICSEARCH_URL}" = "x" ]; then
+  el_url="https://elasticsearch:9200"
+else
+  el_url="${ELASTICSEARCH_URL}"
+fi
+
+# Run as user "elasticsearch" if the command is "elasticsearch"
```
jesuslinares: Explain.

elasticsearch/config/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
...
+if [ "$1" = 'elasticsearch' -a "$(id -u)" = '0' ]; then
+	set -- su-exec eulasticsearch "$@"
+	ES_JAVA_OPTS="-Des.network.host=$NETWORK_HOST -Des.logger.level=$LOG_LEVEL -Xms$HEAP_SIZE -Xmx$HEAP_SIZE"  "$@" &
```
jesuslinares: Where is the HEAP_SIZE variable?. Use default values if it is empty.

elasticsearch/config/entrypoint.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+set -m
+
+# Add elasticsearch as command if needed
```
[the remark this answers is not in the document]
SitoRBJ: It is certainly unnecessary, it is replaced by set -e so that the script aborts when it gets output responses other than 0.

elasticsearch/config/entrypoint.sh
SitoRBJ: In this section is unnecessary, this entry belongs to the script that updates the settings.

elasticsearch/config/entrypoint.sh
SitoRBJ: This is done to run the Elasticsearch CMD. It is not the practice performed by Elasticsearch but emulates it. 

We have decided to change the entrypoint so that it is in tune with the official Elasticsearch entrypoint, performing the tasks we need as they do.

SitoRBJ:
Hello team,

To perform good practices as done in other repositories, we have adapted the Elasticsearch **Dockerfile** to launch an Elasticsearch CMD.

The **entrypoint.sh** contains the options to execute the command independently of the user who launches it. If it is root the command generates problems, that's why, when it is root, a different execution is contemplated specifying the user with UID=1000.

This way of proceeding is obtained from the Elasticsearch repository:

https://github.com/elastic/elasticsearch-docker/blob/6.5.4/build/elasticsearch/bin/docker-entrypoint.sh

Best regards,

Alfonso Ruiz-Bravo

PR #111, opened 2019-02-05
Adding the option to disable some xpack features

LuisGi93:
Hi team,
this PR solves - https://github.com/wazuh/wazuh-saas/issues/41. It adds the capacity of enabling and disabling xpack features mainly at Kibana but also disable by default xpack machinelearning from the elasticsearch image. All these changes happen when the container is created and it is possible to enable and disable xpacks features using environment variables. For example:
```yml
  kibana:
    build: ./kibana
    hostname: kibana
    restart: always
#    ports:
#      - "5601:5601"
    environment:
      - ELASTICSEARCH_URL=http://elasticsearch:9200
      - XPACK_DEVTOOLS=true
      - XPACK_CANVAS=false
      - XPACK_LOGS=true
```
The docker-compose that can be seen above enables the devtools and logs plugin  at Kibana and disables the canvas plugin. Right now by default devtools, canvas, infraestructure, machinelearning, logs and apm are disabled by defautl. When a wazuh-kibana  container is created it will have the following appearance:

image

A battery of tests were done:
- All the docker images were deleted, then all the images were built and containers were created with  the default settings.  The following was checked:
  - no errors were found in the logs.
  -  was able to access the Wazuh app at Kibana. 
  - was checked that some alerts were received from the manager. 
  - also was checked that the kibana.yml and the elasticserach.yml had the correct values. 
- The containers were launched again with the same settings. No duplicates values were found at kibana.yml and the elasticserach.yml.  They had the same values.

- The values of the  environment variables were inverted. The containers were created again  and the correct values were found at the .yml configuration files. Then the containers were stopped and started again. No duplicate values, no bad indentations and again the correct values were found at the .yml files.

- Also an hybrid test was done with some variables set to true and other to false and checked that the corresponding values were set at the .yml files.

The following environment variables were added at Kibana:
- XPACK_DEVTOOLS: enables/disables the devtools plugin.
- XPACK_CANVAS= enables/disables the canvas plugin.
 - XPACK_LOGS= enables/disables the logs plugins.
 - XPACK_INFRA= enables/disables the infraestructure plugins.
 - XPACK_ML= enables/disables the Machine Learning plugin.
 - XPACK_APM= enables/disables the devtools plugins.
They need to have the values true for enabling a feature and false to disable it.

### DIFF
--- a/elasticsearch/config/load_settings.sh
+++ b/elasticsearch/config/load_settings.sh
@@ -84,9 +84,9 @@ then
   done
 else
   echo "
-  #xpack features
-  xpack.ml.enabled: $XPACK_ML
-  " >> $elasticsearch_config_file
+#xpack features
+xpack.ml.enabled: $XPACK_ML
+ " >> $elasticsearch_config_file
 fi
 
 echo "Elasticsearch is ready."
--- a/elasticsearch/config/load_settings.sh
+++ b/elasticsearch/config/load_settings.sh
@@ -68,25 +68,5 @@ curl -XPUT "$el_url/_cluster/settings" -H 'Content-Type: application/json' -d'
 }
 '
 
-#Disabling xpack features
-
-elasticsearch_config_file="/usr/share/elasticsearch/config/elasticsearch.yml"
-if grep -Fq  "#xpack features" "$elasticsearch_config_file";
-then 
-  declare -A CONFIG_MAP=(
-  [xpack.ml.enabled]=$XPACK_ML
-  )
-  for i in "${!CONFIG_MAP[@]}"
-  do
-    if [ "${CONFIG_MAP[$i]}" != "" ]; then
-      sed -i 's/.'"$i"'.*/'"$i"': '"${CONFIG_MAP[$i]}"'/' $elasticsearch_config_file
-    fi
-  done
-else
-  echo "
-#xpack features
-xpack.ml.enabled: $XPACK_ML
- " >> $elasticsearch_config_file
-fi
 
 echo "Elasticsearch is ready."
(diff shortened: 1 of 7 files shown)


elasticsearch/config/load_settings.sh
```diff
@@ -68,4 +68,25 @@ curl -XPUT "$el_url/_cluster/settings" -H 'Content-Type: application/json' -d'
...
+      sed -i 's/.'"$i"'.*/'"$i"': '"${CONFIG_MAP[$i]}"'/' $elasticsearch_config_file
+    fi
+  done
+else
+  echo "
+#xpack features
+xpack.ml.enabled: $XPACK_ML
+ " >> $elasticsearch_config_file
+fi
+
```
SitoRBJ: The content of this section runs once Elasticsearch has been started. If we want Elasticsearch to take this configuration it should be implemented before it starts. For example, we could create a second bash script that is launched in the foreground before this script and before the start of Elasticsearch.

kibana/config/entrypoint.sh
```diff
@@ -23,4 +23,7 @@ sleep 5
 
 ./kibana_settings.sh &
 
+
+./xpack_config.sh &
+
```
SitoRBJ: This script must be executed before KIbana starts. In theory it does, but it is not necessary to launch it in the background.

LuisGi93:
Hello @SitoRBJ,
thanks for you for warning me about these bugs! I've made the changes to solve them. I'm going to set all environment variables to true. Also, thanks for noticing the "Kibana plugin optimization". This PR adds **the capacity to disable some xpacs features** at Kibana and at Elasticsearch. I have set all the environment variables to "true" **but** in case of setting and environment variable to disable some xpack feature **it will cause** that when the Kibana container starts it will do the **"Optimizing and caching browser bundles"**.

Issue #50, opened 2018-08-29
Review Kibana configuration when using X-Pack security


JuanjiJG:
Review this Wazuh mailing list post: https://groups.google.com/forum/#!topic/wazuh/pCzCwZG1rRw

Add options when the user is using X-Pack security on Elasticsearch, Logstash and Kibana.

Regards,
Juanjo

manuasir:
Done here https://github.com/wazuh/wazuh-docker/pull/119

PR #119, opened 2019-02-12
Added support for secure HTTPS xPack requests

manuasir:
Hello team,

This PR adds the ability to perform requests to the Elastic API in the case of xPack enabled.
Related issue: #50 

Cheers


elasticsearch/config/load_settings.sh
```diff
@@ -15,8 +15,13 @@ else
...
 fi
 
+if [ "x${ELASTICSEARCH_USERNAME}" = "x"]; then
+  auth=""
+else
+  auth="--user ${ELASTICSEARCH_USERNAME}:${ELASTICSEARCH_PASSWORD}"
+fi
 
-until curl -XGET $el_url; do
+until curl ${auth} -XGET $el_url; do
```
LuisGi93: Better to follow the same style as the rest of the document for specifying the auth credentials: 
`curl -XGET $el_url  ${auth} `

LuisGi93:
Hi team,
Today a first test of this PR was done. I checked that the changes proposes don't affect the normal behavior of an installation without xPack enabled. Did not find any problem.

elasticsearch/config/load_settings.sh
```diff
@@ -15,8 +15,13 @@ else
   wazuh_url="${WAZUH_API_URL}"
 fi
 
+if [ "x${ELASTICSEARCH_USERNAME}" = "x"]; then
```
LuisGi93: An space need to be added after the "x":
[ "x${ELASTICSEARCH_USERNAME}" = "x" ];
otherwise the following error will occur:
`./load_settings.sh: line 18: [: missing `]'`

Issue #123, opened 2019-03-07
Elasticsearch container doesnt work using docker run


Phandora:
Hi,

I have been working with docker in a Centos7. I got the Wazuh repository following the steps on the installation guide for docker: git clone https://github.com/wazuh/wazuh-docker.git -b 3.8.2_6.5.4 --single-branch.

I needed to install a plugin (repository-s3 from aws) and run a script into elasticsearch container, so I added some lines at the end of load_settings.sh file. I used docker run instead of docker compose and because of that some environmental variables (like ELASTICSEARCH_URL) weren't properly set up and program kept running in a loop waiting for Elastic to be available.

Regards.

PR #124, opened 2019-03-12
Fixing docker run loop at the elasticsearch container

LuisGi93:
This PR fix - https://github.com/wazuh/wazuh-docker/issues/123. A change was made at the url where the elasticsearch configuration is set. Instead of "elasticsearch" that was set up by docker-compose "localhost| is used because when the container is run outside of docker-compose it will enter in a infinite loop because it cant resolve "elasticsearch".

### DIFF
--- a/elasticsearch/config/load_settings.sh
+++ b/elasticsearch/config/load_settings.sh
@@ -4,7 +4,7 @@
 set -e
 
 if [ "x${ELASTICSEARCH_URL}" = "x" ]; then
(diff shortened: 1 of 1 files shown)


Phandora:
I have been testing this PR adding some lines at the end of load_settings.sh file. 

I used `docker build -t <tag>` to create the image and `docker run -d <image_id>` to run the container.

Elasticsearch was available after a while and the lines I added were running smoothly . Everything worked fine.

PR #138, opened 2019-03-26
fix logstash heap size limit

ruffy91:
LS_HEAP_SIZE does not work for logstash versions >= 5.0.0 use LS_JAVA_OPTS instead


ruffy91:
Initial Heap Size has to be set too. I will open a fixed new PR

PR #139, opened 2019-03-26
Use LS_JAVA_OPTS instead of old LS_HEAP_SIZE

ruffy91:
LS_HEAP_SIZE does not work for logstash versions >= 5.0.0 use LS_JAVA_OPTS instead.


manuasir:
Thanks @ruffy91 !

PR #151, opened 2019-04-17
Changing the original Wazuh docker image to allow adding code in entrypoint.sh

Phandora:
Hi team!

This PR changes the official Wazuh docker image to facilitate the process of adding code in the `entrypoint.sh` without using commands like _sed_ or _echo_.

The changes are as follows:
- **00-wazuh.sh** contains the same code from the older `entrypoint.sh` but the call to the init service.
- **entrypoint-scripts folder** is created in order to contain every configuration script.
(e.g. 00-wazuh.sh, 10-cloud-backups.sh)
- **entrypoint.sh** runs every _.sh_ script located in entrypoint-scripts folder in lexicographical order, and finally it calls the init service.

Best regards,
Mayte Ariza


wazuh/config/entrypoint.sh
```diff
@@ -1,154 +1,12 @@
 #!/bin/bash
 # Wazuh App Copyright (C) 2019 Wazuh Inc. (License GPLv2)
 
-#
-# OSSEC container bootstrap. See the README for information of the environment
-# variables expected by this script.
-#
+# It will run every .sh script located in entrypoint-scripts folder in lexicographical order
+for script in `ls /entrypoint-scripts/*.sh | sort -n`; do
+  bash "$script" -H
```
jesuslinares: What means the "-H" argument?.

jesuslinares:
```-H
Enable ! style history substitution. This option is on by default when the shell is interactive.```

Phandora: I have removed "-H" argument

PR #232, opened 2019-08-24
Feature/add executable bit

marco2704:
There were a few `chmod +x` commands in the wazuh, kibana and elasticsearch Dockerfiles. Those commands are not longer needed since the executable bite has been added to be the files. Honored thanks to [core.fileMode](https://git-scm.com/docs/git-config#Documentation/git-config.txt-corefileMode) config variable.

There are other bunch of `chmod 755` command that would be also removed by adding the executable bit too. But I was not sure whether in the future the permission may change.

We can take Postgres [docker-entrypoint.sh](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh) image as an example of this approach.


jm404:
Hi @marco2704,

As far as I know, the `core.fileMode` objective is to keep the execution bit in the scripts but that would imply modifying them to add the execution bit by default. I'm not comfortable with the idea of having scripts with execution bit by default. 

From the security perspective I think it's better to modify the permissions inside the container, not the host.

Anyway we will test it and update you as soon as possible.

Thanks for contributing, it's really appreciated

Best regards,

Jose

PR #261, opened 2019-10-03
Allow using the hostname as node_name for managers

JPLachance:
Hello!

To simplify Wazuh's deployment in Kubernetes, we need the ability to set the node_name dynamically.

I tried to add the support for that in https://github.com/wazuh/wazuh/pull/1611 and https://github.com/wazuh/wazuh/pull/2006, but it was reverted and https://github.com/wazuh/wazuh/issues/2249 is being pushed back at every release 🙂 

So here is the idea:
- Add a simple sed in the Wazuh manager configuration script to replace "to_be_replaced_by_hostname" by the hostname in ossec.conf

I'm running this in production. It works very well and I wanted to share it!

Have a nice day!


wazuh/config/00-wazuh.sh
```diff
@@ -104,6 +104,17 @@ function ossec_shutdown(){
...
+##############################################################################
+# Allow users to set the container hostname as <node_name> dynamically on
+# container start.
+#
+# To use this:
+# 1. Create your own ossec.conf file
+# 2. In your ossec.conf file, set to_be_replaced_by_hostname as your node_name
+# 3. Mount your custom ossec.conf file at $WAZUH_CONFIG_MOUNT/etc/ossec.conf
+##############################################################################
+sed -i 's/<node_name>to_be_replaced_by_hostname<\/node_name>/<node_name>'"${HOSTNAME}"'<\/node_name>/g' /var/ossec/etc/ossec.conf
```
jm404: Hi @JPLachance ,

Let me suggest a little change here, please consider changing `/var/ossec/etc/ossec.conf` to `"${WAZUH_INSTALLATION_PATH}/etc/ossec.conf"` as the default installation path of Wazuh can be changed.

Thanks for your contribution, it's highly appreciated

Best regards,

Jose

PR #293, opened 2020-01-10
Fix S3 plugin installation in Elasticsearch

AnthonySendra:
Hello,

This PR fixes two problems I had with the installation of the S3 plugin.
- The first one is related to [ARG before FROM](https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact) which is not taken into account in the rest of the Dockerfile.
- The second one was related to `${bin/elasticsearch-plugin install --batch S3_PLUGIN_URL}` which did nothing


manuasir:
Thanks for collaborating @AnthonySendra, we'll take a look at this so it can be shipped ASAP.

PR #399, opened 2020-10-29
Set wazuh-wui as default user for the app

xr09:
This PR changes the default user for the kibana app to `wazuh-wui` since it's the one with `rus_as` support.


Zenidd (approved):
LGTM !

PR #400, opened 2020-11-13
Update Dockerfile

nu11secur1ty:
Docker must known who i-node is the parent
( **Docker must known who i-node is the parent**)
# Correct way =)
COPY /config/wazuh.repo /etc/yum.repos.d/wazuh.repo


xr09:
Our usage of the `COPY` command is compliant with Dockers docs: https://docs.docker.com/engine/reference/builder/#copy

There is no difference between `COPY config/wazuh.repo` and `COPY /config/wazuh.repo`

More on this: https://github.com/moby/moby/issues/4592#issuecomment-282702580

nu11secur1ty:
Actually everything is ok =) my mistake sorry

PR #427, opened 2021-01-12
Update generate-self-signed-cert.sh

wireglitch:
Modify permissions of self-generated certs to allow Kibana docker instance the ability to read the file on start.


xr09:
Hi @wireglitch, 

Are you running Ubuntu 20.0.4?

I mention because I've been debugging the cert issue on Ubuntu Focal and so far it all points to permissions of the Docker process.

You might want to take a look at these:

- https://linuxhandbook.com/docker-permission-denied/
- https://forum.snapcraft.io/t/docker-fails-with-permission-denied-inside-containers/17602/16

xr09:
We'll proceed to close the PR since this is a very specific situation on Ubuntu 20.04 env. Allowing all users on the system to read a certificate shouldn't be done lightly.

PR #435, opened 2021-02-03
Re-enable entrypoint scripts

### LINKED ISSUE #433: Unique node_name for wazuh-worker container
Providing a unique name for the wazuh-worker `<node_name>` in the configuration file would allow an arbitrary number of nodes to be created with docker-compose's `--scale` option.
For example by modifying the ossec.conf file with the commands:
```
id=$(hostname -i | tr -d ";.";)
sed 's/<node_name>.*<\/node_name>/<node_name>WAZUH_WORKER-'${id}'<\/node_name>/g' $DIRECTORY/etc/ossec.conf  -i
```
Would append the node's IP to the node_name (without periods) and make it unique.
Allowing the user to specify the number of nodes with the command:
```
docker-compose -f production-cluster.yml up --scale wazuh-worker=3
```

In previous releases, for example 3.9.3_7.2.0 which is featured in [this blogpost](https://wazuh.com/blog/auto-scalable-wazuh-cluster-with-docker-compose/) this would be possible thanks to the use of `/entrypoints-scripts/` which is no longer available.

xr09:
This PR fixes #433 

## Changes:

- Re-enable entrypoint scripts (these were lost since 4.0 when we migrated to s6-overlay)

### DIFF
--- a/wazuh-odfe/config/etc/cont-init.d/2-manager
+++ b/wazuh-odfe/config/etc/cont-init.d/2-manager
@@ -102,12 +102,25 @@ EOF
   fi
 }
 
+function_entrypoint_scripts() {
+  # It will run every .sh script located in entrypoint-scripts folder in lexicographical order
+  if [ -d "/entrypoint-scripts/" ]
+  then
+    for script in `ls /entrypoint-scripts/*.sh | sort -n`; do
+      bash "$script"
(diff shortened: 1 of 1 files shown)


Zenidd (approved):
LGTM !

PR #449, opened 2021-03-03
Modified to be more suitable for production deployment

lostmaniac:
Increase elasticsearch persistent storage to ensure that data is not lost after restart, and modify the directory to the current directory of docker-compose. Easy to migrate


lostmaniac:
Increase elasticsearch persistent storage to ensure that data is not lost after restart, and modify the directory to the current directory of docker-compose. Easy to migrate

PR #470, opened 2021-04-22
Bumped to 4.1.5

alberpilot:
Bump 4.1.5


VERSION
```diff
@@ -1,2 +1,2 @@
-WAZUH-DOCKER_VERSION="4.1.4"
+WAZUH-DOCKER_VERSION="4.1.5"
 REVISION="41100"
```
VictorMorenoJimenez, line 2:
```suggestion
REVISION="40114"
```

VictorMorenoJimenez (changes requested):

VERSION
VictorMorenoJimenez: https://github.com/wazuh/wazuh/blob/v4.1.5/src/REVISION

VictorMorenoJimenez (approved):

PR #482, opened 2021-06-08
Added FILEBEAT_METRICS_ENABLED env var 

maikcaru:
Added FILEBEAT_METRICS_ENABLED env var which is used to control the value of the logging.metrics.enabled setting in filebeat.yml


teddytpc1:
Hi @maikcaru.
Sorry for the delay in this review.
As the original code was updated, we will need the following to be able to merge this PR:
1. Update your branch with the content of branch `4.3`.
2. The default value for `FILEBEAT_METRICS_ENABLED` must be `false`.

PR #519, opened 2021-09-23
Change tail argument to prevent log rotation issues

fhielpos:
Hello team,

This PR aims to fix an issue where the container stops showing logs after the `ossec.log` file gets rotated.

I have replaced the `tail -f` command with `tail -F` to prevent that:

```shell
       -f, --follow[={name|descriptor}]
              output appended data as the file grows;

              an absent option argument means 'descriptor'

       -F     same as --follow=name --retry
```

Regards,
Franco Hielpos


alberpilot (approved):
LGTM

vcerenu (approved):
LGTM

Issue #550, opened 2021-11-11
Consider auto-restarting worker containers once they have been synchronized


fhielpos:
Hello team,

In Kubernetes environments where worker pods can be automatically deployed, a worker container should be automatically restarted after a synchronization so custom rules and decoders can take place.

## Description

Wazuh managers need to be restarted in order to new rules and decoders take place after that have been added. In some environments, workers are automatically deployed and they should not need additional interaction to work properly.

## Possible solutions

We should have an init script to automatically restart the `wazuh-manager` service if the container is a `worker` inside a cluster.

This restart should only take place the first time the container is started.

Regards,
Franco Hielpos

vcerenu:
Added the function_restart_worker function  inside the _wazuh-odfe/config/etc/2-manager_ file , which checks if the Wazuh Manager cluster node is a worker and restarts the Manager services if It's true

```
function_restart_worker() {
  # Extract what type of node is configured in the ossec.conf file
  node_type=`grep node_type /var/ossec/etc/ossec.conf | sed 's/<\/\?[^>]\+>//g'`
  # If the node type is Worker, the wazuh-manager service is restarted for correct registration in the cluster
  if [ $node_type == "worker" ] 
  then
    /var/ossec/bin/wazuh-control restart
  fi
}
```

alberpilot:
I proceed to close this issue. Although it works, the problem is not at the deployment level, the Cluster nodes must be able to hot reload pushed configurations, so we will open a new issue in the main repository.
Regards,
Alberto R

PR #562, opened 2021-12-15
Restart worker manager

vcerenu:
This PR adds a function to restart the Wazuh Manager worker nodes, at the start of the container, for a correct registration
Related Issue #550


teddytpc1 (approved):
LGTM. It is working properly on K8s.

alberpilot:
I procced to close this PR. Although it works, the problem is not at the deployment level, the Cluster nodes must be able to hot reload pushed configurations, so we will open a new issue in the main repository. 
Regards, 
Alberto R

PR #600, opened 2022-03-29
Replace Service name and volumes

vcerenu:
This PR changes the name of the Wazuh Indexer service in the deploy as single node and the name of the volumes related to ossec by wazuh.

Also fixes the migration document from Wazuh 4.2 to 4.3


alberpilot (approved):
LGTM

Issue #511, opened 2021-09-06
Vulnerabilities found on wazuh-odfe image 


MiguelCasaresRobles:
Hello team,

I'm opening this issue because we found some vulnerabilities in the image we are using for `wazuh-odfe.` Find attached here the report.

[wazuh-odfe-scan.json.txt](https://github.com/wazuh/wazuh-docker/files/7116823/wazuh-odfe-scan.json.txt)

We should review them and use an image without vulnerabilities to deploy Wazuh in docker. 

It would be a good idea to remove those packages from the OS not used for Wazuh too.

Regards,

Miguel Casares

vcerenu:
Clair was configured to be able to perform the vulnerability test of the Docker Hub images currently published in the Wazuh repository
```
vcerenu@vcerenu-VirtualBox:~/Repositorios/test-vulnerability/clair$ sudo docker-compose exec clairctl clairctl analyze -l wazuh/wazuh-odfe

Image: /wazuh/wazuh-odfe:latest
 16 layers found
 
  ➜ Analysis [aea1bb27539d] found 0 vulnerabilities.
  ➜ Analysis [a7fca3b9ba2a] found 0 vulnerabilities.
  ➜ Analysis [c26be7fb2650] found 0 vulnerabilities.
  ➜ Analysis [7c1d56602ae4] found 0 vulnerabilities.
  ➜ Analysis [c5c1ea80e70a] found 0 vulnerabilities.
  ➜ Analysis [cf1168b466d2] found 0 vulnerabilities.
  ➜ Analysis [520d0d160129] found 0 vulnerabilities.
  ➜ Analysis [5ac4186f4935] found 0 vulnerabilities.
  ➜ Analysis [29b8faa06275] found 0 vulnerabilities.
  ➜ Analysis [1a7b9fe8aea1] found 0 vulnerabilities.
  ➜ Analysis [314c54583718] found 0 vulnerabilities.
  ➜ Analysis [aadece73ad6b] found 0 vulnerabilities.
  ➜ Analysis [8c7b3cf52339] found 0 vulnerabilities.
  ➜ Analysis [1f7fe1f30bf1] found 0 vulnerabilities.
  ➜ Analysis [512abcaf21ca] found 0 vulnerabilities.
  ➜ Analysis [ad264c3e1db7] found 0 vulnerabilities.
 
cleaning temporary local repository: remove /tmp: permission denied
[...]

PR #615, opened 2022-04-12
Base Change for Wazuh Manager Docker Image

### LINKED ISSUE #510: Adding Alpine Linux docker images as base image for Wazuh-docker deployments
Hello team,

Currently, our Wazuh docker deployment is using centos7 as a base image. However, it will be useful to use Alpine Linux as a possibility for that and leverage Alpine utilities for docker deployments: https://hub.docker.com/_/alpine

By doing so, we will be able to use:

- Lightweight, fast and stable GNU/Linux distribution suitable for containers (5MB)
- Well-optimized regarding vulnerabilities 

We can release this image for only the Wazuh managers by default for those users that are not using Filebeat.

Regards,

Miguel Casares

vcerenu:
This PR is created to modify the current base image from Centos 7 to Ubuntu Focal, which is lighter, contains less CVEs and is still under official support.
Closes #510  #511


teddytpc1 (dismissed):
LGTM

build-docker-images/wazuh-manager/Dockerfile
```diff
@@ -1,27 +1,22 @@
 # Wazuh Docker Copyright (C) 2021 Wazuh Inc. (License GPLv2)
-FROM centos:7
+FROM ubuntu:focal
 
+ARG WAZUH_VERSION=4.3.0
+ARG TEMPLATE_VERSION="master"
```
alberpilot, line 5: Why master? This could be dangerous if we start working on Elastic Common Schema and we modify the template.

alberpilot (changes requested):

build-docker-images/wazuh-manager/Dockerfile
vcerenu: It is a variable that was assigned that way in the previous version, I try to eliminate it and see the deploy behavior

vcerenu: the variable was modified by the value of the branch used in the version

alberpilot (approved):

PR #637, opened 2022-05-12
Change dashboard port

vcerenu:
This PR changes the port on which the Wazuh dashboard starts due to permission problems of the owner user, who cannot use ports below 1024 in some systems


alberpilot (approved):
LGTM

PR #809, opened 2023-04-05
Update 1-config-filebeat

ghaem51:
if remove a single quotation from '$INDEXER_URL' then we can pass the multi-indexer address with env: INDEXER_URL=https://indxer1.wazuh:9200,https://indxer2.wazuh:9200,https://indxer3.wazuh:9200 and its good example for multi-node docker-compose config


ghaem51:
Did no one check this?
@davidkarlsen @jfut @sultanovich @francobep

ghaem51:
Hi everyone,

I noticed that my previous comment regarding setting multiple Wazuh indexers using INDEXER_URL was not accepted. Could someone clarify why this approach wasn't considered valid?

Additionally, if we still want to configure multiple Wazuh indexers for a Wazuh worker/master setup, what would be the recommended solution? Is there an alternative method or a specific reason why modifying $INDEXER_URL is not suitable?

Looking forward to your insights. Thanks!
@c-bordon

PR #913, opened 2023-07-30
Remove unnecessarily exposed ports in single-node docker-compose.yml …

colinmollenhour:
As far as I can tell it is not necessary to expose these ports to have a working Wazuh server.
The dashboard communicates with the indexer and manager via the internal docker network. The exposed ports are only needed to view the dashboard (443) and for agents to connect remotely (1514,1515,514).
Am I missing something?

### DIFF
--- a/single-node/docker-compose.yml
+++ b/single-node/docker-compose.yml
@@ -10,7 +10,6 @@ services:
       - "1514:1514"
       - "1515:1515"
       - "514:514/udp"
-      - "55000:55000"
     environment:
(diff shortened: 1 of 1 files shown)


matiasmercado-ar:
In "exposed ports" in the documentation https://documentation.wazuh.com/current/deployment-options/docker/wazuh-container.html, I found that the port 55000 is for Wazuh API and 9200 is for Wazuh indexer HTTPS

Issue #1563, opened 2024-10-08
filebeat.conf file is not updated if file already exists in /etc/filebeat/filebeat.yml


mattjamesaus:
We came across an issue with the filebeat.yml file not getting updated when bumping the wazuh manager version. In our case we had a version of https://github.com/wazuh/wazuh-docker/blob/master/build-docker-images/wazuh-manager/config/filebeat.yml#L26 that didn't have the seccomp updates (4.6) and when the container booted up to 4.9.x it didn't update the filebeat.yml in our persistent volume.

My thinking is that https://github.com/wazuh/wazuh-docker/blob/205983317f3e09ba460d0d30ba6e23835b442258/build-docker-images/wazuh-manager/config/permanent_data.env#L18 should probably contain /etc/filebeat/filebeat.yml so that in the event the docker container has an update to filebeat.yml the volume mount (kube or compose) would have it's filebeat.yml file updated to match the newer version in the container.

Enaraque:
Hi,
Thank you very much for your suggestion, you're absolutely right.
We will implement this improvement in the next open release (4.14.0).

Best regards!

Enaraque:
## Modifying `filebeat.yml` after adding it to `PERMANENT_DATA_EXCP`

I've added the `/etc/filebeat/filebeat.yml` file to the `PERMANENT_DATA_EXCP` list and built the images with this configuration. The version used was `v4.13.0-alpha1`.

<details>
<summary>Building process</summary>

```console
$ sudo bash ./build-docker-images//build-images.sh -v 4.13.0 -r 1 -d alpha1 -f 0.4
Compose can now delegate builds to bake for better performance.
 To do so, set COMPOSE_BAKE=true.
[+] Building 155.8s (98/98) FINISHED                                                                                                                                                          docker:default
 => [wazuh.manager internal] load build definition from Dockerfile                                                                                                                                      0.0s
 => => transferring dockerfile: 2.38kB                                                                                                                                                                  0.0s
 => [wazuh.agent internal] load build definition from Dockerfile                                                                                                                                        0.0s
 => => transferring dockerfile: 1.23kB                                                                                                                                                                  0.0s
 => [wazuh.indexer internal] load build definition from Dockerfile                                                                                                                                      0.0s
 => => transferring dockerfile: 2.76kB                                                                                                                                                                  0.0s
 => [wazuh.dashboard internal] load metadata for docker.io/library/amazonlinux:2023                                                                                                                     1.4s
[...]

PR #1898, opened 2025-06-20
Add filebeat config file in the `PERMANENT_DATA_EXCP` list

Enaraque:
# Description

When upgrading between different Wazuh images, the `/etc/filebeat/` path was included in the list of permanent data, so the file `/etc/filebeat/filebeat.yml` was never updated.
This PR changes that behavior by adding this file to the `PERMANENT_DATA_EXCP` list so that when upgrading to a new version, the file is updated as well.

## Testing 🧪 

You can see how the upgrade from one version to another was tested with these changes in this comment:

- https://github.com/wazuh/wazuh-docker/issues/1563#issuecomment-2991037102

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 
 ### Changed
 
+- Add filebeat config file in the PERMANENT_DATA_EXCP list ([#1898](https://github.com/wazuh/wazuh-docker/pull/1898))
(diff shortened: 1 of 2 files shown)


teddytpc1 (approved):

wroe1583:
It seems that the same problem occurs with /etc/filebeat/wazuh-template.json. After the upgrade, the old template is still used, and indices are being created without the proper mapping. Does anyone know if this issue has been addressed?

Issue #1677, opened 2025-01-14
MVP - Change build image process


vcerenu:
## Description

According to the modifications to be made in the Wazuh package repositories, the installation method of the binaries of each component in its respective image must be modified. We are currently using RPM packages through yum, so we should change the method of obtaining the package and installing it in the build process. This process has to contemplate the creation of both production and pre-release and development images.

It is also necessary to add to the installation logic the creation of both amd64 and arm64 images, so it is necessary to add this logic in the image build.

This build process must be carried out in a Github Actions workflow, which must contain the logic to push images in each release to the Wazuh Docker Hub repository as well as to be able to push images from development branches, pushing the images to private repositories in ECR.

Issue #923, opened 2023-08-07
Official arm64 image support?


alessiodam:
I would like to know if it was possible of making the wazuh/wazuh docker image also available for arm64, the main point is to get it working on a pi at home.

aswath-betaflux:
Same Ask from me aswell. Since we are running on ARM based bode it will be great to include support for ARM. 

I was trying to build by cloning but lot of packages has hardcoded to 64 bit in the docker build process. 

Wish Team wazuh provide us both X86 and Arm based docker images.

tkbstudios:
Yeah hoped too :)

tkbstudios:
Any updates for the arm version?

jeankhawand:
Well I tried to get it working there are two key packages which I didn't find for the arm indexer and dashboard but for the manager it's possible to port it for arm
@vcerenu @tkbstudios

tkbstudios:
I approved the changes but I know it won't make it mergable lol thanks man I will finally be able to host it on a Pi 4/5 👍

Hussienfahmy:
Is there any updates on the packages or any workaround to work fully on arm?

collse:
I am also highly interested on arm to run on Oracle cloud

code2319:
can someone help with https://github.com/wazuh/wazuh-docker/pull/950 ?

```
❯ ./build-images.sh
ERROR [wazuh.manager  4/15] RUN curl -L -O https://packages.wazuh.com/4.x/apt/pool/main/w/wazuh-manager/wazuh-manager_4.8.0-1_arm64.deb &&  1.3s
------
 > [wazuh.manager  4/15] RUN curl -L -O https://packages.wazuh.com/4.x/apt/pool/main/w/wazuh-manager/wazuh-manager_4.8.0-1_arm64.deb &&    dpkg -i wazuh-manager_4.8.0-1_arm64.deb && rm -f wazuh-manager_4.8.0-1_arm64.deb:
0.271   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
0.271                                  Dload  Upload   Total   Spent    Left  Speed
100   336    0   336    0     0    333      0 --:--:--  0:00:01 --:--:--   333
1.296 dpkg-deb: error: 'wazuh-manager_4.8.0-1_arm64.deb' is not a Debian format archive
1.299 dpkg: error processing archive wazuh-manager_4.8.0-1_arm64.deb (--install):
[...]

PR #1946, opened 2025-07-25
Add arm64 platform support to image build

yoursw:


gelbphoenix:
### This could possibly close following Issues:
Closes: #923
Closes: #1677

yoursw:
@vcerenu - would you happen to have availability to check this - it seems a simple fix. Thanks

vs4vijay:
Much needed.

vcerenu:
Hello @yoursw 

Thank you very much for your contribution!!

Could you provide evidence of how this change has worked in your environment? You can test it with an already published release to have pre-built ARM packages available.
According to our plan, we need to modify the s6-overlay dependency. We haven't yet tested the ARM version of it, and we need to be able to manage services within the Wazuh Manager Docker image. We plan to work on this issue in our roadmap, but we don't have a date/release for this change yet. If you can generate working images with this change, I can submit it for approval and add it more quickly to our development process.

pmharris77:
> Hello @pmharris77 
> 
> 
> 
> We have developed image builds for amd64 and arm64 for Wazuh version 5.0.0, which is currently under development. We will upload the images for both architectures to Docker Hub for this version.
> 
> 
> 
> Image building is currently possible, so you can build images on an arm64 host using the steps outlined in our documentation:
> 
> https://documentation.wazuh.com/current/deployment-options/docker/build-docker-images-locally.html
> 
> 
> 
> The changes made in the pull request are insufficient, as they require a configured environment with buildx and QEMU to generate the images. Furthermore, they are not suitable for generating them locally because local registries do not support manifests for multiple architectures. Therefore, the build will only be able to build the host architecture, allowing for the creation of local images that can then be deployed on the same host.

Ah perfect, thanks @vcerenu, I had no idea that v5.0.0 would be the first to have multi-arch images available. Is there any rough timescales on when it will be ready? Trying to figure out if I should wait, or build locally myself.

teddytpc1:
The ARM images are being included for 5.0.0. This PR will be closed.

PR #2435, opened 2026-05-28
Feature/multi node sni haproxy

eazeved:


eazeved:
ok

vcerenu:
Hello @eazeved 

Thank you so much for your contribution!

At this time, we won't be able to merge these changes due to some incompatibilities with the deployment and the purpose of our repository.

First, the change you're making is to the main branch, where we're working on version 5.0.0, which is currently under development. Therefore, the version change isn't correct, as many of the changes made are not compatible with Wazuh 4.x versions.
Second, the current Wazuh 5.0.0 deployment is being done as an example deployment, using the fewest possible customizations and simplifying the deployment as much as possible. This way, users can use this initial deployment as a basis for building their own deployment according to their specific business and infrastructure needs.

I recommend that you save the changes you're sending us so you can customize the deployment as needed when we have Wazuh 5.0.0 in production.

PR #2581, opened 2026-08-13
Change Codebuild runners to Github runners

vcerenu:
Related Issue https://github.com/wazuh/wazuh-automation/issues/3665

### DIFF
--- a/.github/workflows/4_build_and_push_images.yml
+++ b/.github/workflows/4_build_and_push_images.yml
@@ -6,7 +6,6 @@ on:
     inputs:
       image_tag:
         description: 'Docker image tag'
-        default: '4.14.9'
         required: true
       docker_reference:
         description: 'wazuh-docker reference'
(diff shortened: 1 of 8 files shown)


jotamolas (changes requested):
Two issues in the new AWS/S3 handling introduced here (same pattern in #2580 and #2581, both touch 4_pr_check.yml):

1. `S3_ARTIFACTS_PATH` in "Temporarily save Wazuh Docker images" now uses `S3://...` (uppercase). `aws s3 cp` requires lowercase `s3://` — verified locally, uppercase fails with `ParamValidation: Invalid argument type`. The download steps further down in this same file already use the correct lowercase `s3://` prefix — please match that.

2. The new `Configure AWS credentials` steps (aws-actions/configure-aws-credentials@v4, OIDC role-to-assume) need `permissions: id-token: write` at the workflow or job level. Neither is declared here, so the OIDC token request will fail at runtime. See wazuh-installation-assistant#972 (same migration, same issue #3665) for the correct pattern:

   permissions:
     id-token: write
     contents: read

Both are one-line fixes but will break CI as-is — requesting changes.

jotamolas (approved):
Confirmed the fix — thanks for the quick turnaround. Approving.

PR #2586, opened 2026-08-18
Adapt manager certificate deployment to unified layout

### LINKED ISSUE #2584: Adapt certificate deployment to the unified manager certificate layout (root:wazuh-manager 0640, dir 1770)
## Description

The Wazuh manager has unified all its certificates under `/var/wazuh-manager/etc/certs` and now applies ownership by role. This deployment repository provisions the manager's certificates, so it must be updated to match the new layout, names, ownership and modes.

Changes in `wazuh/wazuh` (design: wazuh/wazuh#38281; bug: wazuh/wazuh#38276; PR: wazuh/wazuh#38278; installation assistant: wazuh/wazuh-installation-assistant#970).

## New layout (all relative to the install dir, under `etc/certs`)

- **Indexer trust material** — `root-ca.pem`, `indexer-connector.pem`, `indexer-connector-key.pem`:
  - Owner `root:wazuh-manager`, mode `0640`. The manager reads it after dropping privileges but must not be able to replace its own trust anchor.
- **Self-generated server certificates** — `authd.*`, `remoted.*`, `apid.*`:
  - Owner `wazuh-manager:wazuh-manager`, mode `0640` (the daemons generate these themselves).
- **Directory** `etc/certs`:
  - Owner `root:wazuh-manager`, mode `1770` (sticky), to support the mixed ownership.

## What changes for this repo

1. **Naming.** The manager's indexer client certificate is now `indexer-connector.pem` / `indexer-connector-key.pem` (was `manager.pem` / `manager-key.pem`). `root-ca.pem` is unchanged.
2. **Ownership and modes.** Where the deployment currently sets the manager certs as `wazuh-manager:wazuh-manager` (or `root:root`) with owner-only modes, it must instead leave the indexer material as `root:wazuh-manager 0640` and the directory as `root:wazuh-manager 1770`, without a recursive `chown -R` that would clobber the self-generated certs.

## Why

After the manager drops privileges to `wazuh-manager`, its daemons and framework read the indexer trust material as that user, so it must be group-readable (`0640`), not owner-only (`400`/`500`). Owner-only modes break startup (the engine reports the certificate as missing — see wazuh/wazuh#38276).

## Tasks

- [ ] Analyze the impact on this repository's certificate deployment
- [ ] Update the certificate placement, names, ownership and modes
- [ ] Update documentation if applicable
- [ ] Test a deployment and provide evidence of the files' ownership and modes

## References

- wazuh/wazuh#38281 (design / proposal)
- wazuh/wazuh#38276 (manager startup failure)
- wazuh/wazuh#38278 (implementation PR, merged in 5.0.0-https)
- wazuh/wazuh-installation-assistant#970 (installation assistant)

jesusph04:
## Problem 

The Wazuh manager unified all its certificates under `etc/certs` (relative to the install dir, i.e. `/var/wazuh-manager/etc/certs`) and now applies ownership by role (design: wazuh/wazuh#38281, implementation: wazuh/wazuh#38278). This repository provisions the manager's certificates and needed to be updated to match:

- **Indexer trust material** (`root-ca.pem`, `indexer-connector.pem`, `indexer-connector-key.pem`): owner `root:wazuh-manager`, mode `0640`. The manager reads this after dropping privileges but must not be able to replace its own trust anchor.
- **Self-generated server certificates** (`authd.*`, `remoted.*`, `apid.*`): owner `wazuh-manager:wazuh-manager`, mode `0640` (generated by the daemons themselves — unaffected by this PR).
- **Directory `etc/certs`**: owner `root:wazuh-manager`, mode `1770` (sticky), to support the mixed ownership above.

Two things needed to change here:
1. **Naming**: the manager's indexer client certificate is now `indexer-connector.pem`/`indexer-connector-key.pem` (was `manager.pem`/`manager-key.pem`).
2. **Ownership and modes**: wherever this repo set the manager's indexer-trust certs to `wazuh-manager:wazuh-manager`/`root:root` with owner-only modes, it now leaves them as `root:wazuh-manager 0640`, and the `etc/certs` directory as `root:wazuh-manager 1770`.

## Changes

- **`build-docker-images/wazuh-manager/Dockerfile`**: `etc/certs` directory creation changed from `chown wazuh-manager:wazuh-manager` + `chmod 500` to `chown root:wazuh-manager` + `chmod 1770`. This is a multi-stage Dockerfile; only the `builder` stage needed the change, since the final stage uses `COPY --from=builder`, which preserves ownership.

- **`single-node/docker-compose.yml`** and **`multi-node/docker-compose.yml`** (both `wazuh.master` and `wazuh.worker` services): renamed the bind-mount destination inside the manager container from `manager.pem`/`manager-key.pem` to `indexer-connector.pem`/`indexer-connector-key.pem`. The source file names on the host side (`wazuh.manager.pem`, etc. — the indexer cert generator's own naming convention) are unrelated to this issue and were left untouched.

- **`tools/utils/deployment/certificates-conf.sh`**: for the `MANAGER_NODES` and `root-ca` blocks only, changed host-side ownership/mode from `${WAZUH_UID}:${WAZUH_GID} 400` to `0:${WAZUH_GID} 640`. Verified this is safe for the shared `root-ca.pem` file (bind-mounted into manager, indexer, and dashboard): all three images use the same GID (`101`) for their respective service users, even though the group names differ, so changing the owning user to `root` doesn't affect group-readability for indexer or dashboard. The `INDEXER_NODES` and `DASHBOARD_NODES` blocks were **not** touched — those certs belong to their own services and are out of scope here. No recursive `chown -R` was introduced or changed in a way that could affect `authd.*`/`remoted.*`/`apid.*`; those are generated and owned by the manager daemons themselves and are untouched by this repo's tooling entirely.

## Validation

Deployed a real single-node stack on AWS (via `workflow_dispatch` on `5_check_integration_tools.yml`, building the manager image from this branch) and confirmed:

**Functional health** — 12 integration tests run, 11 passed, 1 skipped (unrelated: a revision-check test with no revision commands configured):

Status: PASS 🟢
Total Tests: 12 | Passed: 11 | Failed: 0 | Skipped: 1

**Explicit ownership/mode evidence** (`ls -ld` / `ls -la` on `/var/wazuh-manager/etc/certs` inside the running manager container):

```
drwxrwx--T 2 root wazuh-manager 4096 Aug 18 13:08 /var/wazuh-manager/etc/certs

-rw-r----- 1 wazuh-manager wazuh-manager 1704 apid-key.pem
-rw-r----- 1 wazuh-manager wazuh-manager 1233 apid.pem
-rw-r----- 1 wazuh-manager wazuh-manager 1704 authd-key.pem
-rw-r----- 1 wazuh-manager wazuh-manager 1261 authd.pem
-rw-r----- 1 root wazuh-manager 1708 indexer-connector-key.pem
-rw-r----- 1 root wazuh-manager 1298 indexer-connector.pem
-rw-r----- 1 wazuh-manager wazuh-manager 1704 remoted-key.pem
-rw-r----- 1 wazuh-manager wazuh-manager 1261 remoted.pem
-rw-r----- 1 root wazuh-manager 1204 root-ca.pem
```
Every value matches the layout specified in #2584:
| Item | Expected | Observed |
|---|---|---|
| `etc/certs` dir | `root:wazuh-manager 1770` | ✅ |
| `root-ca.pem` | `root:wazuh-manager 0640` | ✅ |
| `indexer-connector.pem` / `-key.pem` | `root:wazuh-manager 0640` | ✅ |
| `authd.*` / `remoted.*` / `apid.*` | `wazuh-manager:wazuh-manager 0640` | ✅ |

This evidence was captured by temporarily adding an explicit `certificates-conf.sh --priv` call plus an `ls`-based verification step to the integration test workflow, on this branch only; both were removed before requesting review (see side finding below on why `--priv` had to be added explicitly).

### DIFF
--- a/.github/workflows/5_check_integration_tools.yml
+++ b/.github/workflows/5_check_integration_tools.yml
@@ -691,6 +691,14 @@ jobs:
             sudo docker compose ps
           "
 
+      - name: Apply certificate ownership/permissions (--priv) for #2584 evidence
+        run: |
+          DEPLOYMENT="${{ matrix.deployment_type }}"
+          ssh ${{ env.SSH_OPTS }} "${{ env.REMOTE }}" "
+            cd /tmp/wazuh-docker/${DEPLOYMENT}
+            sudo bash /tmp/wazuh-docker/tools/utils/deployment/certificates-conf.sh --priv
+          "
+
       - name: Verify certificate ownership and modes (evidence for #2584)
         run: |
           DEPLOYMENT="${{ matrix.deployment_type }}"
--- a/.github/workflows/5_check_integration_tools.yml
+++ b/.github/workflows/5_check_integration_tools.yml
@@ -691,6 +691,23 @@ jobs:
             sudo docker compose ps
           "
 
+      - name: Verify certificate ownership and modes (evidence for #2584)
+        run: |
+          DEPLOYMENT="${{ matrix.deployment_type }}"
+          ssh ${{ env.SSH_OPTS }} "${{ env.REMOTE }}" "
+            cd /tmp/wazuh-docker/${DEPLOYMENT}
+            echo '=== Manager certificate ownership and modes (issue #2584 evidence) ==='
+            for SERVICE in wazuh.manager wazuh.master wazuh.worker; do
+              if sudo docker compose ps --services 2>/dev/null | grep -qx \"\$SERVICE\"; then
+                echo \"--- Service: \$SERVICE ---\"
+                echo '[Directory etc/certs]'
+                sudo docker compose exec -T \"\$SERVICE\" ls -ld /var/wazuh-manager/etc/certs
+                echo '[Certificate files]'
+                sudo docker compose exec -T \"\$SERVICE\" ls -la /var/wazuh-manager/etc/certs
+              fi
+            done
+          "
+
       # -----------------------------------------------------------------------
       # Run integration tests
       # -----------------------------------------------------------------------
--- a/.github/workflows/5_check_integration_tools.yml
+++ b/.github/workflows/5_check_integration_tools.yml
@@ -691,31 +691,6 @@ jobs:
             sudo docker compose ps
           "
 
-      - name: Apply certificate ownership/permissions (--priv) for #2584 evidence
-        run: |
-          DEPLOYMENT="${{ matrix.deployment_type }}"
-          ssh ${{ env.SSH_OPTS }} "${{ env.REMOTE }}" "
-            cd /tmp/wazuh-docker/${DEPLOYMENT}
-            sudo bash /tmp/wazuh-docker/tools/utils/deployment/certificates-conf.sh --priv
-          "
-
-      - name: Verify certificate ownership and modes (evidence for #2584)
-        run: |
-          DEPLOYMENT="${{ matrix.deployment_type }}"
-          ssh ${{ env.SSH_OPTS }} "${{ env.REMOTE }}" "
-            cd /tmp/wazuh-docker/${DEPLOYMENT}
-            echo '=== Manager certificate ownership and modes (issue #2584 evidence) ==='
-            for SERVICE in wazuh.manager wazuh.master wazuh.worker; do
-              if sudo docker compose ps --services 2>/dev/null | grep -qx \"\$SERVICE\"; then
-                echo \"--- Service: \$SERVICE ---\"
-                echo '[Directory etc/certs]'
-                sudo docker compose exec -T \"\$SERVICE\" ls -ld /var/wazuh-manager/etc/certs
-                echo '[Certificate files]'
-                sudo docker compose exec -T \"\$SERVICE\" ls -la /var/wazuh-manager/etc/certs
-              fi
-            done
-          "
-
       # -----------------------------------------------------------------------
       # Run integration tests
       # -----------------------------------------------------------------------
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 
 | Issue | Comment |
 | - | - |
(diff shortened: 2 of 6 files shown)


jotamolas (approved):
Logic checks out. The scoping is careful — `INDEXER_NODES`/`DASHBOARD_NODES` blocks in `certificates-conf.sh` correctly left untouched, and the shared-GID reasoning for `root-ca.pem` (bind-mounted into manager/indexer/dashboard, all using GID 101) is explicitly verified in the description, not just assumed. The real `ls -ld`/`ls -la` evidence in the PR body is solid, concrete proof this works end-to-end.

Two non-blocking notes:

1. Traced `.github/workflows/5_check_integration_tools.yml` (the standard automated integration test, same one used for `/test-docker-single`) — it calls `certificates-conf.sh --cert --copy` **without `--priv`**. So the standing CI never exercises the chown/chmod logic this PR changes, before or after. The `ls` evidence above is real but one-off (captured by temporarily adding `--priv` to the workflow, per the description) — nothing protects this going forward if the permission logic regresses later. Might be worth permanently adding `--priv` to that workflow step.
2. Related to the above: the description says "see side finding below on why `--priv` had to be added explicitly" but there's no such section in the current PR body — looks like a paragraph got dropped when cleaning up before requesting review. Worth restoring or removing the dangling reference.
3. Same shape as a note I left on wazuh-kubernetes#1612: the Dockerfile's `chown root:wazuh-manager`/`chmod 1770` on `/var/wazuh-manager/etc/certs` only takes effect when the named volume (`master-wazuh-etc`) is created fresh from the image — an existing deployment upgrading in place keeps the directory's old owner/mode (the individual bind-mounted cert files still get renamed/re-permissioned correctly every start, that part isn't affected).

Approving.

PR #2602, opened 2026-08-24
Install diffutils in Wazuh agent image

### LINKED ISSUE #2594: `report_changes` cannot work: no `diff` binary in the agent image
|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-latest | wazuh-agent Docker image, wazuh-syscheckd error reporting | Manager/Agent | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), single-node stack |

## Description

With `report_changes="yes"` on a monitored directory, every file modification logs:

```
wazuh-syscheckd: ERROR: (6714): Command diff output an error
```

The FIM event itself is still produced, and the state index still updates the size and hash, so only the content-diff part of the feature is lost.

## Root cause

The image has no `diff` on the path:

```
docker exec <agent> sh -c 'ls /usr/bin/diff'   ->  No such file or directory
docker exec <agent> sh -c 'which diff'          ->  which: command not found
```

So a documented FIM option fails on every use in the container image. The error message never names the missing binary, which makes it hard to diagnose from the log alone.

## Steps to reproduce

```bash
docker exec <agent> sh -c 'mkdir -p /testfim'
# add to the agent's ossec.conf:
#   <directories realtime="yes" check_all="yes" report_changes="yes">/testfim</directories>
docker restart <agent>
docker exec <agent> sh -c 'echo a > /testfim/f; sleep 3; echo b >> /testfim/f'
docker logs <agent> 2>&1 | grep 6714
```

dev-rehaann:
## Description

Install `diffutils` in the Wazuh agent runtime image so FIM `report_changes` can execute `/usr/bin/diff`.

Update the 5.0.0 changelog with the fix.

Closes #2594

## Testing

- Rebuilt the PR branch from the current `5.0.0` head so the diff is limited to the runtime dependency and changelog entry.
- Validated the changelog entry against the repository workflow format and ran `git diff --check`.
- Previously built the complete Wazuh agent image with the 5.0.0-beta4 RPM and verified the final runtime image contains `/usr/bin/diff`, `diffutils-3.8-1.amzn2023.0.2.x86_64`, and GNU diff 3.8.
- The package installation step was also confirmed on amd64 and arm64 during review.

A fresh complete build from the rebuilt branch could not be run locally because the Docker Desktop backend fails before starting the engine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 | Issue | Comment |
 | - | - |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
+| [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
 | [#2590](https://github.com/wazuh/wazuh-docker/issues/2590) | Restore `WAZUH_MANAGER_PORT` support in the Wazuh agent image and document the enrollment variables removed in 5.0.0 |
 | [#2578](https://github.com/wazuh/wazuh-docker/issues/2578) | Report skipped bumps in the repository bumper workflow |
 | [#2584](https://github.com/wazuh/wazuh-docker/issues/2584) | Adapt certificate deployment to the unified manager certificate layout (root:wazuh-manager 0640, dir 1770) |
--- a/build-docker-images/wazuh-agent/Dockerfile
+++ b/build-docker-images/wazuh-agent/Dockerfile
@@ -77,7 +77,7 @@ ARG WAZUH_GID=101
 RUN rm /bin/sh && ln -s /bin/bash /bin/sh
 
 # Install only runtime dependencies
-RUN dnf install procps shadow-utils -y && \
+RUN dnf install diffutils procps shadow-utils -y && \
     dnf clean all && \
     getent group wazuh || groupadd -r -g ${WAZUH_GID} wazuh && \
     getent passwd wazuh || useradd --system \


jesusph04:
Confirmed the actual `diffutils` install step works correctly on both
architectures (see `stage-1 3/8` in the build log — diffutils
3.8-1.amzn2023.0.2 installs cleanly, adding ~1.4 MB installed size).

However, the full build fails afterwards in an unrelated step (the
`builder` stage downloading the `wazuh-agent` RPM itself returns a
truncated 342-byte response — looks like an invalid/expired presigned
URL). This is likely because the branch is based on `main`, which is
already at `5.1.0-alpha0` — a version that probably has no published
dev artifacts yet.

patriiciacoboss (changes requested):
Since the `5.0.0` branch already has published artifacts, would it make sense to rebuild this PR against `5.0.0` instead of `main`? That would let the pipeline fetch a real RPM via the presigned URL and complete build, confirming `report_changes` actually produces the diff inside the final container — rather than leaving that validation blocked until `5.1.0-alpha0` has published artifacts.

jesusph04:
However, the full build fails afterwards in an unrelated step (the
`builder` stage downloading the `wazuh-agent` RPM itself returns a
truncated 342-byte response — looks like an invalid/expired presigned
URL). This is likely because the branch is based on `main`, which is
already at `5.1.0-alpha0` — a version that probably has no published
dev artifacts yet.

dev-rehaann:
That makes sense. I checked the branch history and confirmed the `diffutils` change applies unchanged to `5.0.0`. The current head is based on `main`; simply changing the PR base would add 29 unrelated files to the diff, so the branch needs to be rebuilt from `5.0.0` instead.

The package-install layer has been validated on both amd64 and arm64. A complete build from the rebuilt branch is still needed to verify `report_changes` in the final agent image.

dev-rehaann:
Rebuilt the branch from the current `5.0.0` head and retargeted the PR. Commit `3410713d` keeps the diff to the agent runtime dependency and the 5.0.0 changelog entry.

The changelog format and `git diff --check` pass. A fresh complete image build is still unavailable locally because Docker Desktop fails before starting its engine; the earlier full 5.0.0-beta4 runtime verification and the amd64/arm64 package-install results remain valid. The branch is now ready for a complete build using the published 5.0.0 artifacts.

vcerenu:
Hi @dev-rehaann 

I wanted to ask if you could redo the commit associated with the branch and verify it. Our merge policy requires that commits be verified before they can be merged.

dev-rehaann:
Thanks for the reminder. I rebuilt the branch on the current 5.0.0 head and recreated the change as GitHub-verified commit a1948860. This also removes the conflict introduced when 5.0.0 advanced.

The PR is now mergeable, one commit ahead and zero behind, with the diff still limited to the diffutils runtime dependency and changelog entry. git diff --check and the changelog-format validation pass. Docker's engine is still unavailable locally, so I could not run a fresh complete image build; the earlier amd64/arm64 package-install validation remains unchanged.

jotamolas:
The changes have been merged in this pull request: https://github.com/wazuh/wazuh-docker/pull/2620. Thank you, @dev-rehaann, for your contribution.

PR #2621, opened 2026-09-03
Change cluster key assigment

vcerenu:
Related issue https://github.com/wazuh/internal-devel-requests/issues/6070


mjbonacci (dismissed):
LGTM

mjbonacci (dismissed):
Updated:
 - WAZUH_CLUSTER_KEY=${WAZUH_CLUSTER_KEY:-}
 - wazuh-cluster-key:/wazuh-cluster-key

mjbonacci (dismissed):
- **Escape-hatch fix verified**: ran the actual `sed_escape_replacement()` + the final `sed` substitution from `0-wazuh-init` (not a reimplementation) inside Amazon Linux 2023, against the exact 3 vectors from the original report:
    - `key-normal-1234` → written unchanged
    - `abc&def` → written unchanged (previously corrupted)
    - `abc|def` → written unchanged (previously corrupted)
    - `abc\def` → written unchanged
  - **Build order verified**: the cluster key is cleared in the same `RUN` layer, before `permanent_data.sh` runs, so the snapshot restored into a fresh `etc` volume also ships the empty key.
  - **CI**: re-ran `/test-docker` against this HEAD. Same 3 failures as the previous run, in both single-node and multi-node (port 9200 not listening / indexer health endpoint / indexer connectivity). Confirmed unrelated to this PR: 9200 stopped being published to the host as of #6058, and the shared `integration-test-module` suite hasn't caught up with that — not a regression from this change.

LGTM

jotamolas:
# Review

Conflicts are resolved as of the latest commits (`b60d870f` + `88a4fe50`), `mergeable` is back to `MERGEABLE`. One thing worth cleaning up before merging, though:

Bringing in `5.0.0` in the last merge commit pulled in `escape_repl()` (added by #2631, already on `5.0.0`), which has the exact same body as this branch's own `sed_escape_replacement()`:

```bash
escape_repl() { printf '%s' "$1" | sed -e 's/[\\&|]/\\&/g'; }
sed_escape_replacement() { printf '%s' "$1" | sed -e 's/[\\&|]/\\&/g'; }
```

Since they live in different parts of the file, git merged them cleanly with no conflict marker, but we ended up with two identical functions. Could you drop `sed_escape_replacement()` and have the `<key>` substitution call `escape_repl()` instead, same as the other `sed` calls in this file (`node_name`, `node_type`, `bind_addr`)?

Also: the two integration test runs (single-node and multi-node) from 2026-09-11 are still showing FAIL (3/12) and haven't been re-triggered since. Worth an `/test-docker` once the above is fixed.

jotamolas (approved):
LGTM

PR #2640, opened 2026-09-07
Default WAZUH_CLUSTER_BIND_ADDR to 0.0.0.0 so the published manager image actually starts

### LINKED ISSUE #2634: The published manager image never starts: its own init writes an empty cluster bind address that its schema rejects
## Summary

`docker run` on the published 5.0.0 manager image, with no arguments and no environment, produces a container that reports `running` and in which no Wazuh daemon is alive.

```
   docker run -d --name probe public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest

   container status                running
   wazuh-manager daemons alive     0
   configuration errors in the log 3
```

The log says why:

```
   (1244): Invalid configuration at '/cluster/bind_addr': does not satisfy 'minLength'
           [schema /properties/cluster/properties/bind_addr].
   wazuh-manager.conf: Configuration error. Exiting
   wazuh-manager-control: ERROR: (1244): Invalid configuration at '/cluster/bind_addr' ...
```

The image ships a valid value and its own entrypoint destroys it.

## The image is correct until the init script edits it

Straight out of the image, before anything runs:

```
   <cluster>
     <name>wazuh</name>
     <node_name>node01</node_name>
     <node_type>master</node_type>
     <key>fb3ee99772cf1620fe021b4921f0edd2</key>
     <port>1516</port>
     <bind_addr>127.0.0.1</bind_addr>          <- valid
     ...
```

and in the container after `/etc/cont-init.d/0-wazuh-init` has run:

```
     <bind_addr></bind_addr>                    <- empty, and the schema requires minLength
```

## One line, and the one beside it shows the intended pattern

`/etc/cont-init.d/0-wazuh-init` substitutes both bind addresses the same way, but defaults only one of them:

```
   line 293   WAZUH_REMOTE_BIND_ADDR="${WAZUH_REMOTE_BIND_ADDR:-0.0.0.0}"     <- defaulted
   line 295   sed -i "/<https>/,/<\/https>/ s|<bind_addr>.*</bind_addr>|<bind_addr>$WAZUH_REMOTE_BIND_ADDR</bind_addr>|"

   line 316   sed -i "/<cluster>/,/<\/cluster>/ s|<bind_addr>.*</bind_addr>|<bind_addr>$WAZUH_CLUSTER_BIND_ADDR</bind_addr>|"
```

`WAZUH_CLUSTER_BIND_ADDR` is never given a default anywhere in the script, so on any deployment that does not set it the substitution expands to nothing and writes an empty element. Every other value in the same block is defaulted from the shipped configuration a few lines earlier, `WAZUH_CLUSTER_KEY` among them.

## Setting the variable is the whole difference

Same image, same command, one variable added:

```
                                   without          with WAZUH_CLUSTER_BIND_ADDR=0.0.0.0
   container status                running          running
   wazuh-manager daemons alive      0                2
   configuration errors             3                0
```

## The container does not report the failure

`docker ps` shows the container `Up`, and it stays up: the entrypoint runs the init, runs the start script, and then `tail -F` on the log file, which keeps PID 1 alive whether or not the manager started. So an orchestrator sees a healthy container, the API port is exposed and refuses connections, and nothing restarts it. The image declares no `HEALTHCHECK`.

That is the part that turns a start-up failure into a silent one: the only signal is inside the log the container is tailing.

## Reproducing

```
docker run -d --name probe313 public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest
sleep 45
docker inspect -f '{{.State.Status}}' probe313        # running
docker exec probe313 sh -c 'ps -eo comm | grep -c wazuh-manager'   # 0
docker logs probe313 | grep -E '1244|Configuration error'
docker rm -f probe313
```

and the control:

```
docker run -d --name probe313b -e WAZUH_CLUSTER_BIND_ADDR=0.0.0.0 \
  public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest
sleep 45
docker exec probe313b sh -c 'ps -eo comm | grep -c wazuh-manager'  # 2
docker rm -f probe313b
```

The two lines in the init script:

```
docker run --rm --entrypoint /bin/bash \
  public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest \
  -c "grep -n 'BIND_ADDR' /etc/cont-init.d/0-wazuh-init"
```

## Suggested fix

Give the variable a default beside the one it sits next to:

```
   WAZUH_CLUSTER_BIND_ADDR="${WAZUH_CLUSTER_BIND_ADDR:-127.0.0.1}"
```

`127.0.0.1` keeps the shipped behaviour for a single node; `0.0.0.0` is the value a clustered deployment wants and is what `WAZUH_REMOTE_BIND_ADDR` already defaults to. Either is better than empty, and the same script should not substitute an unset variable into a required element in the first place.

Separately, the entrypoint should not go on to `tail -F` a log for a manager that failed to start. Exiting non-zero would let Docker's restart policy and any orchestrator see the failure, and a `HEALTHCHECK` that looks for a running `wazuh-manager-analysisd` would catch the rest.

## Environment

- `public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest`, pulled 2026-09-04, image built 2026-09-04T03:52:58Z.
- `VERSION.json`: `{"version": "5.0.0", "stage": "rc1", "commit": "7990fc0"}`.
- Reproduced with no volumes, no network and no environment beyond the image's own.

## Related

- [314](BUG_314-the-manager-cannot-be-pointed-at-a-plain-http-indexer-and-the-tls-it-insists-on-needs-three-certificates-the-image-never-creates.md) and [315](BUG_315-no-agent-can-enrol-against-the-published-manager-because-it-signs-its-own-enrolment-certificate-and-the-agent-demands-full-verification.md) are the next two things that stop the same stack coming up, once this one is worked around.
- [259](BUG_259-the-only-directory-the-shipped-compose-does-not-persist-is-the-one-holding-the-detection-content.md) is the other defect in how this image is packaged for deployment.

jesusph04:
## Description 

`0-wazuh-init` substitutes `WAZUH_CLUSTER_BIND_ADDR` into `<cluster><bind_addr>`, but unlike its sibling `WAZUH_REMOTE_BIND_ADDR` (already defaulted to `0.0.0.0` at line 293), it never gets a default
anywhere in the script. A plain `docker run` with no environment variables leaves it unset, so the substitution writes an empty `<bind_addr>`, which the manager's config schema rejects (`minLength` violation) — every daemon fails to start.

The failure is silent: the entrypoint runs `tail -F` on the log afterward regardless of whether the manager started, so the container stays `Up` and looks healthy to any orchestrator.

## Fix

Same one-line pattern already used for the sibling variable:
```bash
WAZUH_CLUSTER_BIND_ADDR="${WAZUH_CLUSTER_BIND_ADDR:-0.0.0.0}"
```
Added right before the `<bind_addr>` substitution, matching the issue's suggested fix exactly.

## Validation

Reproduced the exact scenario from the issue's "Reproducing" section,patching a local build of `wazuh-manager:5.0.0-latest` with this change:

| | Before (official image) | After (this fix) |
|---|---|---|
| Container status | `running` | `running` |
| `wazuh-manager` daemons alive | 0 | 2 |
| Error 1244 in logs | present | absent |

```
$ docker logs probe_before 2>&1 | grep -E '1244|Configuration error' (1244): Invalid configuration at '/cluster/bind_addr': does not satisfy 'minLength' [schema /properties/cluster/properties/bind_addr].
wazuh-manager.conf: Configuration error. Exiting

$ docker logs probe_after 2>&1 | grep -E '1244|Configuration error'
(no output)
```

## Additional validation — full image build from source

Also validated with a complete build from the Dockerfile (not just a
patch over the published image), using real RPM artifacts via
`artifact_urls.yaml` + `build-images.sh -c wazuh-manager -v 5.0.0`:

```
$ docker run -d --name probe_real_build wazuh/wazuh-manager:5.0.0
$ sleep 45 && docker inspect -f '{{.State.Status}}' probe_real_build
running
$ docker exec probe_real_build sh -c 'ps -eo comm | grep -c wazuh-manager'
2
$ docker logs probe_real_build 2>&1 | grep -E '1244|Configuration error'
(no output)
```

Same result as the earlier patched-image validation — 2 daemons alive,
no config errors

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@
 
 | Issue | Comment |
 | - | - |
+| [#2634](https://github.com/wazuh/wazuh-docker/issues/2634) | Default WAZUH_CLUSTER_BIND_ADDR to 0.0.0.0 so the published manager image actually starts |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
 | [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
 | [#2590](https://github.com/wazuh/wazuh-docker/issues/2590) | Restore `WAZUH_MANAGER_PORT` support in the Wazuh agent image and document the enrollment variables removed in 5.0.0 |
--- a/build-docker-images/wazuh-manager/config/etc/cont-init.d/0-wazuh-init
+++ b/build-docker-images/wazuh-manager/config/etc/cont-init.d/0-wazuh-init
@@ -313,6 +313,7 @@ sed -i "/<cluster>/,/<\/cluster>/ s|<key>.*</key>|<key>$WAZUH_CLUSTER_KEY</key>|
 # --------------------------
 # Cluster: bind_addr
 # --------------------------
+WAZUH_CLUSTER_BIND_ADDR="${WAZUH_CLUSTER_BIND_ADDR:-0.0.0.0}"
 sed -i "/<cluster>/,/<\/cluster>/ s|<bind_addr>.*</bind_addr>|<bind_addr>$WAZUH_CLUSTER_BIND_ADDR</bind_addr>|" "$WAZUH_MANAGER_CONF"
 
 # --------------------------


vcerenu (approved):
LGTM

jotamolas (approved):
LGTM

PR #2642, opened 2026-09-07
Fix manager and dashboard healthchecks to correctly detect failures

### LINKED ISSUE #2627: The shipped container healthchecks report healthy for three of the four failures they exist to catch
## Summary

`wazuh-manager-control status` distinguishes four states per daemon and already exits non-zero for the three bad ones. The healthcheck in the shipped `docker-compose.yml` throws that exit code away and greps the human-readable output for the single string `not running`, so a daemon that failed to start, a daemon that refused its configuration, and a `status` invocation that produced no output at all are all reported as healthy. Nothing else notices: the container's entrypoint ignores whether the manager started and then waits on `tail -F` forever, so the container never exits and `restart: always` never fires. In the multi-node compose the worker and the dashboard both wait on `condition: service_healthy` against that same check, so they start against a manager that never came up.

The dashboard healthcheck in the same file has the same shape of defect and its own twist: it omits `curl -f`, so it ignores the HTTP status, and the URL it probes returns 404 on a perfectly healthy dashboard.

## The manager healthcheck

From `single-node/docker-compose.yml` on the 5.0.0 branch of `wazuh-docker`, and identically for `wazuh.master` and (with a `grep -v apid` added) for `wazuh.worker` in `multi-node/docker-compose.yml`:

```yaml
    healthcheck:
      test: [ "CMD-SHELL", "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -q 'not running' && exit 1 || exit 0" ]
      interval: 15s
      timeout: 5s
      retries: 5
      start_period: 60s
```

`status()` in `/var/wazuh-manager/bin/wazuh-manager-control` has four output branches, not two:

```sh
        pstatus ${i};
        if [ $? = 0 ]; then
            if [ -f ${DIR}/var/run/${i}.failed ]; then
                if [ $USE_JSON = true ]; then
                    echo -n '{"daemon":"'${i}'","status":"failed"}'
                elif [ "`cat ${DIR}/var/run/${i}.failed 2>/dev/null`" = "refused" ]; then
                    echo "${i} refused its configuration..."
                else
                    echo "${i} failed to start..."
                fi
            elif [ $USE_JSON = true ]; then
                echo -n '{"daemon":"'${i}'","status":"stopped"}'
            else
                echo "${i} not running..."
            fi
            RETVAL=1
        else
            ...
            echo "${i} is running..."
        fi
```

`RETVAL=1` is set for all three failure branches and the script ends in `exit $RETVAL`, so `wazuh-manager-control status` on its own is already a correct healthcheck. The grep discards it.

The healthcheck is a pure string test, so feeding it each string the code above can emit is an exact measurement of what it detects:

```
   all daemons up                    -> exit 0  HEALTHY
   stopped                           -> exit 1  unhealthy
   failed to start                   -> exit 0  HEALTHY
   refused its configuration         -> exit 0  HEALTHY
   empty output (the script fails)   -> exit 0  HEALTHY
   the binary does not exist         -> exit 0  HEALTHY
```

One of the six is caught. The `refused its configuration` branch is the one that matters most in practice, because it is exactly what `testconfig()` produces when a configuration value is rejected, and the marker file it reads exists for the express purpose of making that state distinguishable. The comment above it says so:

```sh
            # The marker testconfig() leaves on a refused configuration, and the same one the
            # framework reads as 'failed'. Without it a rejected value is indistinguishable from a
            # daemon that was never started.
```

The product went to the trouble of separating "rejected the configuration" from "never started", and the healthcheck collapses both into "healthy".

The last two rows matter because the check is fail-open by construction. Any failure of the pipeline itself, a missing binary, a `status` that dies before printing, an exec that cannot run during a restart, produces no output, so `grep -q` fails, so the `|| exit 0` branch fires and the container is declared healthy. A healthcheck should report the failure it cannot rule out; this one reports success.

## Nothing else catches it either

The image's entrypoint does not check whether the manager started:

```bash
bash /etc/cont-init.d/0-wazuh-init
bash /etc/cont-init.d/1-manager

tail -F /var/wazuh-manager/logs/wazuh-manager.log &
TAIL_PID=$!
...
wait "${TAIL_PID}"
```

and `/etc/cont-init.d/1-manager` ends, at line 99, with a bare

```sh
/var/wazuh-manager/bin/wazuh-manager-control start
```

with no `set -e` and no test of the result. The container therefore stays up whatever happened, `restart: always` never fires because nothing ever exits, and the only remaining signal is the healthcheck, which says healthy.

In `multi-node/docker-compose.yml` that signal is load-bearing:

```
   wazuh.worker    depends_on: wazuh.master    condition: service_healthy
   wazuh.dashboard depends_on: wazuh.master    condition: service_healthy
```

so the worker and the dashboard are released to start against a master whose daemons refused their configuration.

The `grep -v apid` in the worker's variant is dead: `status()` already skips `wazuh-manager-apid` when the node type is not master, so the line it filters is never printed on a worker.

## The dashboard healthcheck, same file

```yaml
      test: [ "CMD", "curl", "-k", "-s", "-o", "/dev/null", "https://localhost:5601/login" ]
```

No `-f`, so curl exits 0 for any response it manages to read, including 500 and 502. Measured against the running dashboard in this deployment, which has no `server.basePath` set:

```
   /login                     the path of the healthcheck  without -f: 404 exit_code=0   with -f: 404 exit_code=22
   /app/does-not-exist-at-all a path that does not exist   without -f: 200 exit_code=0   with -f: 200 exit_code=0
   /api/status                the api of state         without -f: 200 exit_code=0   with -f: 200 exit_code=0
```

Two things at once. The probed path does not exist, `/login` returns 404 and the login page is served from `/app/login`; and because the single-page application answers 200 for any unknown `/app/...` path, the check would pass on a completely arbitrary URL just as well. It cannot distinguish a working dashboard from one that only serves its shell.

The indexer healthcheck three services down the same file shows the authors know the correct shape:

```yaml
      test: [ "CMD-SHELL", "curl -fks https://localhost:9200/_plugins/_security/health | grep -q '\"status\":\"UP\"'" ]
```

`-f` for the status code and a grep for a positive assertion, so both a bad status and an unexpected body fail the check.

## Reproducing

The manager half needs no live failure, since the check is a string test:

```
for s in 'x is running...' 'x not running...' 'x failed to start...' 'x refused its configuration...' ''; do
  printf '%s' "$s" | grep -q 'not running' && echo "$s -> unhealthy" || echo "$s -> HEALTHY"
done
```

The dashboard half, from inside the dashboard container:

```
curl -k -s -o /dev/null -w '%{http_code} exit=' https://localhost:5601/login ; echo $?
curl -fk -s -o /dev/null -w '%{http_code} exit=' https://localhost:5601/login ; echo $?
```

## Suggested fix

The manager healthcheck is already written for you by the script it calls:

```yaml
      test: [ "CMD", "/var/wazuh-manager/bin/wazuh-manager-control", "status" ]
```

which reports unhealthy for stopped, failed and refused alike, and, being a `CMD` rather than a shell pipeline, cannot silently succeed when the command itself fails. If the worker really needs to tolerate a missing API daemon, do that by node type in the script rather than by grepping its prose.

For the dashboard, probe a path that exists and check the status code:

```yaml
      test: [ "CMD-SHELL", "curl -fks https://localhost:5601/api/status | grep -q '\"state\":\"green\"'" ]
```

Independently of the healthchecks, `1-manager` should propagate the result of `wazuh-manager-control start`, and the entrypoint should not begin tailing a log for a manager that never started.

## Environment

- Manager 5.0.0, image `wazuh/wazuh-manager:5.0.0-latest`, compose from the 5.0.0 branch of `wazuh-docker` at `37452c5c` (2026-08-24).
- Dashboard 5.0.0, no `server.basePath` configured.
- The manager container in this deployment carries the compose's own `nofile` and `memlock` ulimits, and `wazuh-manager-control status` currently reports all eight daemons running with exit code 0.

## The one case this healthcheck does catch

For completeness, the `not running` branch is not hypothetical, and it is the branch that fires in
the failure documented in
[096](BUG_096-nothing-restarts-the-manager-api-when-it-dies-and-the-container-stays-up-marked-unhealthy.md):
killing the API's master process leaves `wazuh-manager-control status` printing
`wazuh-manager-apid not running...`, the grep matches, and the container does turn `unhealthy` after
about two minutes. So the check works for a daemon that died and for nothing else. That finding also
shows what the verdict is worth once it is correct: the container stays `running`, `RestartCount`
stays at 0, and nothing acts on the `unhealthy` state.

patriiciacoboss:
## Manager healthcheck
 
**single-node and the cluster master:** `wazuh-manager-control status` already sets the correct exit code for all these failure states on its own. So instead of grepping the text, I now just run the command directly and let Docker read its exit code.
 
**cluster worker:** this one couldn't get the same simple fix. The worker's healthcheck has an extra piece: it filters out any line mentioning `apid` before checking for failures (`grep -v apid`). That's intentional — `apid` is only supposed to run on the master, so on the worker it will always show as "not running", and that's normal, not a failure.
 
If I switched the worker to the raw exit code like I did for the master, I'd lose that filtering. The exit code reflects the state of every daemon at once, `apid` included, so the worker would end up permanently marked `unhealthy` for something that's expected behavior.
 
So for the worker, I kept the text-based grep, but fixed the actual bug: instead of only checking for `not running`, it now checks for all three known failure phrases.
 
## Verification
 
I couldn't spin up a real `5.1.0` stack to test end-to-end — neither the Docker image nor the certificate-generation scripts for that version are published yet. Instead, I tested the exact logic with a stand-in script that reproduces the same outputs `wazuh-manager-control status` can give, and ran the real healthcheck commands against it.
 
**single-node / master**, comparing the old logic against the new one for every possible state:
 
```
=== OLD logic, state=healthy ===
RESULT: exit 0 (healthy)
=== OLD logic, state=not_running ===
RESULT: exit 1 (unhealthy)
=== OLD logic, state=failed_to_start ===
RESULT: exit 0 (healthy)          <- bug: should be unhealthy
=== OLD logic, state=refused_config ===
RESULT: exit 0 (healthy)          <- bug: should be unhealthy
 
=== NEW logic, state=healthy ===
RESULT: exit 0 (healthy)
=== NEW logic, state=not_running ===
RESULT: exit 1 (unhealthy)
=== NEW logic, state=failed_to_start ===
RESULT: exit 1 (unhealthy)        <- fixed
=== NEW logic, state=refused_config ===
RESULT: exit 1 (unhealthy)        <- fixed
```
 
**worker**, confirming the apid exclusion still works and the missed failure states are now caught:
 
```
=== WORKER logic (new), state=healthy (apid reported "not running", as expected) ===
RESULT: exit 0 (healthy)          <- apid exclusion still works correctly
=== WORKER logic (new), state=not_running ===
RESULT: exit 1 (unhealthy)
=== WORKER logic (new), state=failed_to_start ===
RESULT: exit 1 (unhealthy)        <- previously a false healthy
```

## Real container verification
 
Managed to get real `5.0.0` packages after all, so I built the images locally with `build-images.sh --dev` and ran the actual stack with `docker compose up`.
 
**Healthy case** — full stack up, manager running all 7 daemons:
 
```
$ docker ps --format "table {{.Names}}\t{{.Status}}"
NAMES                         STATUS
single-node-wazuh.dashboard   Up 34 seconds (healthy)
single-node-wazuh.manager     Up 44 seconds (healthy)
single-node-wazuh.indexer     Up About a minute (healthy)
 
$ docker exec single-node-wazuh.manager /var/wazuh-manager/bin/wazuh-manager-control status
wazuh-manager-clusterd is running...
wazuh-manager-modulesd is running...
wazuh-manager-remoted is running...
wazuh-manager-analysisd is running...
wazuh-manager-db is running...
wazuh-manager-authd is running...
wazuh-manager-apid is running...
Exit code: 0
```
 
**Failure case** — stopped the daemons for real inside the container, then waited for the healthcheck's `retries: 5` × `interval: 15s` window:
 
```
$ docker exec single-node-wazuh.manager /var/wazuh-manager/bin/wazuh-manager-control stop
Wazuh v5.0.0 Stopped
 
$ docker exec single-node-wazuh.manager /var/wazuh-manager/bin/wazuh-manager-control status
wazuh-manager-clusterd not running...
wazuh-manager-modulesd not running...
[...all 7 not running]
Exit code: 1
 
# ~90s later
$ docker inspect single-node-wazuh.manager --format='{{.State.Health.Status}}'
unhealthy
```
 
Docker correctly picked up the failure via the new healthcheck and marked the container `unhealthy`, on a real manager, not just the stand-in script from the earlier verification.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ### Added
 
-
 | Issue | Comment |
 | - | - |
 | [#2606](https://github.com/wazuh/wazuh-docker/issues/2606) | Support the unified `WAZUH_MANAGER_ENDPOINT` connection URL in the Wazuh agent, alongside `WAZUH_MANAGER_SERVER` and `WAZUH_MANAGER_PORT` |
@@ -15,7 +14,6 @@
 | [#2276](https://github.com/wazuh/wazuh-docker/issues/2276) | Add `--set-as-main` flag support to repository bumper — `wazuh-docker` |
 
 ### Changed
-
 
 | Issue | Comment |
 | - | - |
@@ -75,10 +73,9 @@
 
 ### Fixed
 
-
 | Issue | Comment |
 | - | - |
-| [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Manager healthcheck now uses the command's own exit code instead of grepping for 'not running', catching all failure states |
+| [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Manager healthcheck now catches all failure states, not just 'not running': single-node and the cluster master use the command's own exit code, and the worker checks for all known failure strings instead of just one |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
 | [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
 | [#2590](https://github.com/wazuh/wazuh-docker/issues/2590) | Restore `WAZUH_MANAGER_PORT` support in the Wazuh agent image and document the enrollment variables removed in 5.0.0 |
@@ -95,6 +92,7 @@
 | [#2271](https://github.com/wazuh/wazuh-docker/issues/2271) | Wazuh manager Healthcheck |
 | [#2258](https://github.com/wazuh/wazuh-docker/issues/2258) | Delete WAZUH_AGENT_GROUPS of Wazuh 5.0.0 images build |
 | [#2128](https://github.com/wazuh/wazuh-docker/issues/2128) | Development - DevOps 5.0 adaptation - Docker - Delete lists directory references |
+
 
 ## Prior versions
 - []()
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,6 +78,7 @@
 
 | Issue | Comment |
 | - | - |
+| [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Manager healthcheck now uses the command's own exit code instead of grepping for 'not running', catching all failure states |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
 | [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
 | [#2590](https://github.com/wazuh/wazuh-docker/issues/2590) | Restore `WAZUH_MANAGER_PORT` support in the Wazuh agent image and document the enrollment variables removed in 5.0.0 |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,7 @@
 
 | Issue | Comment |
 | - | - |
-| [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Manager healthcheck now catches all failure states, not just 'not running': single-node and the cluster master use the command's own exit code, and the worker checks for all known failure strings instead of just one |
+| [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Fixed manager and dashboard healthchecks to correctly detect failures instead of always reporting healthy |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
 | [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
 | [#2590](https://github.com/wazuh/wazuh-docker/issues/2590) | Restore `WAZUH_MANAGER_PORT` support in the Wazuh agent image and document the enrollment variables removed in 5.0.0 |
--- a/multi-node/docker-compose.yml
+++ b/multi-node/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       wazuh1.indexer:
         condition: service_healthy
     healthcheck:
-      test: [ "CMD-SHELL", "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -q 'not running' && exit 1 || exit 0" ]
+      test: [ "CMD", "/var/wazuh-manager/bin/wazuh-manager-control", "status" ]
       interval: 15s
       timeout: 5s
       retries: 5
@@ -49,7 +49,7 @@ services:
     container_name: multi-node-wazuh.worker
     restart: always
     healthcheck:
-      test: [ "CMD-SHELL", "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -v apid | grep -q 'not running' && exit 1 || exit 0" ]
+      test: [ "CMD-SHELL", "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -v apid | grep -qE 'not running|failed to start|refused its configuration' && exit 1 || exit 0" ]
       interval: 15s
       timeout: 5s
       retries: 5
--- a/multi-node/docker-compose.yml
+++ b/multi-node/docker-compose.yml
@@ -202,7 +202,7 @@ services:
     container_name: multi-node-wazuh.dashboard
     restart: always
     healthcheck:
-      test: [ "CMD", "curl", "-k", "-s", "-o", "/dev/null", "https://localhost:5601/login" ]
+      test: [ "CMD", "curl", "-f", "-k", "-s", "-o", "/dev/null", "https://localhost:5601/app/login" ]
       interval: 30s
       timeout: 10s
       retries: 3
--- a/single-node/docker-compose.yml
+++ b/single-node/docker-compose.yml
@@ -9,7 +9,7 @@ services:
       wazuh.indexer:
         condition: service_healthy
     healthcheck:
-      test: [ "CMD-SHELL", "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -q 'not running' && exit 1 || exit 0" ]
+      test: [ "CMD", "/var/wazuh-manager/bin/wazuh-manager-control", "status" ]
       interval: 15s
       timeout: 5s
       retries: 5
--- a/single-node/docker-compose.yml
+++ b/single-node/docker-compose.yml
@@ -87,7 +87,7 @@ services:
     container_name: single-node-wazuh.dashboard
     restart: always
     healthcheck:
-      test: [ "CMD", "curl", "-k", "-s", "-o", "/dev/null", "https://localhost:5601/login" ]
+      test: [ "CMD", "curl", "-f", "-k", "-s", "-o", "/dev/null", "https://localhost:5601/app/login" ]
       interval: 30s
       timeout: 10s
       retries: 3


jotamolas (commented):
Manager healthcheck fix looks correct and well validated (both the simulated states and the real stop/unhealthy test).

Per the original issue #2627, could you also fix the dashboard healthcheck in the same PR? Same shape of defect, different lines: `single-node/docker-compose.yml:90` and `multi-node/docker-compose.yml:205` - no `-f`, so curl exits 0 on any response including 4xx/5xx, and it probes `/login` which 404s on a healthy dashboard (the real login page is `/app/login`). The issue's suggested fix:

```yaml
test: [ "CMD-SHELL", "curl -fks https://localhost:5601/api/status | grep -q '"state":"green"'" ]
```

patriiciacoboss:
## Dashboard healthcheck
 
Same shape of defect as the manager one: `curl` had no `-f`, so it exited `0` on any HTTP response, including 4xx/5xx — the healthcheck could never actually fail regardless of what the dashboard returned. It also probed `/login`, which returns `401` on a healthy dashboard (the real login page is `/app/login`).
 
The issue originally suggested checking `/api/status` for `"state":"green"` instead, but I confirmed against a real container that this endpoint requires authentication in this version (`401 Unauthorized` without credentials) — that suggested fix would never succeed even on a healthy dashboard. Used `/app/login` with `-f` instead, which does return `200` without authentication.
 
## Verification
 
Same approach as the manager fix: tested the command manually against a real container first, then confirmed the healthcheck itself (not just the command) correctly transitions the container between `healthy` and `unhealthy`.
 
**Manual command check, healthy case:**
 
```
$ docker exec single-node-wazuh.dashboard curl -f -k -s -o /dev/null https://localhost:5601/app/login
Exit code: 0
```
 
**Manual command check, frozen process (simulated with SIGSTOP):**
 
```
$ docker exec single-node-wazuh.dashboard curl -f -k -s -o /dev/null --max-time 5 https://localhost:5601/app/login
Exit code: 28
```
 
**Real healthcheck, recreated the container so the new definition is actually used:**
 
```
$ docker compose up -d --force-recreate wazuh.dashboard
$ docker ps --format "table {{.Names}}\t{{.Status}}"
NAMES                         STATUS
single-node-wazuh.dashboard   Up About a minute (healthy)
```
 
**Then froze the process again and let Docker's own healthcheck catch it:**
 
```
$ docker kill --signal=SIGSTOP single-node-wazuh.dashboard
# waited past retries: 3 x interval: 30s
 
$ docker inspect single-node-wazuh.dashboard --format='{{json .State.Health}}'
{
    "Status": "unhealthy",
    "FailingStreak": 10,
    "Log": [
        {
            "Output": "Health check exceeded timeout (10s)"
        },
        {
            "Output": "Health check exceeded timeout (10s)"
        },
        {
            "Output": "Health check exceeded timeout (10s)"
        },
        {
            "Output": "Health check exceeded timeout (10s)"
        },
        {
            "Output": "Health check exceeded timeout (10s)"
        }
    ]
}
```
 
**Resumed the process and confirmed recovery:**
 
```
$ docker kill --signal=SIGCONT single-node-wazuh.dashboard
$ docker ps --format "table {{.Names}}\t{{.Status}}"
NAMES                         STATUS
single-node-wazuh.dashboard   Up 13 minutes (healthy)
```
 
Docker correctly picked up the failure via the new healthcheck and marked the container `unhealthy`, and correctly recovered it back to `healthy` once the process resumed — on a real dashboard, not just a manually-run command.

jotamolas (dismissed):
Independently re-verified both fixes against a real running stack (single-node, certs generated with the real wazuh-certs-tool.sh, images built from this branch).

**Manager**: healthy -> stopped the daemons for real (`wazuh-manager-control stop`) -> Docker marked the container `unhealthy` after 5 consecutive failed checks (exit code 1 each time, all 8 daemons reporting `not running`).

**Dashboard**: healthy -> froze the process (`SIGSTOP`) -> `unhealthy` after 3 consecutive failed checks -> `SIGCONT` -> recovered to `healthy`.

Both transitions match exactly what's described in the PR's own verification section. Approving.

jotamolas (approved):
After resolving the CHANGELOG conflict, LGTM.

PR #2646, opened 2026-09-08
Escape special sed replacement characters so passwords and keys with &, \, or the delimiter are no longer corrupted or silently discarded

### LINKED ISSUE #2631: A password containing an ampersand, a pipe or a backslash is silently rewritten by the config injection sed
## Summary

The manager, dashboard and agent images all write configuration values into their config files with `sed -i "s|pattern|literal ${value}|"`, without escaping the value. Three characters are special on the right-hand side of a `sed` substitution and none is handled: `&` expands to the whole matched text, `\` starts an escape, and `|` ends the substitution because it is the delimiter. A secret containing any of them is silently replaced by something else, or the substitution fails and the shipped default is left in place. In every case the script exits 0 and the deployment starts with credentials the operator did not choose.

## The dashboard API password

`/wazuh_dashboard_config.sh` injects the Wazuh API credentials into the `wazuh_core.hosts` block:

```bash
    sed -i "/^wazuh_core.hosts:/,/^[^ ]/ {
        s|url:.*|url: $WAZUH_API_URL|
        s|port:.*|port: $API_PORT|
        s|username:.*|username: $API_USERNAME|
        s|password:.*|password: $API_PASSWORD|
        s|run_as:.*|run_as: $RUN_AS|
    }" "$DASHBOARD_CONFIG_FILE"
```

Running the image's own script with a range of passwords and reading back what landed in `opensearch_dashboards.yml`:

```
  alfanumerica                   -> intacta
  with point and script              -> intacta                 ("MyS3cr37P450r.*-")
  with ampersand                  -> SENT 'abc&def'   WRITTEN 'abcpassword: wazuh-wuidef'  rc=0
  with slash vertical             -> SENT 'abc|def'   WRITTEN 'wazuh-wui'                  rc=1
                                    sed: -e expression #1, char 183: unknown option to `s'
  with slash invertida            -> SENT 'abc\def'   WRITTEN 'abcdef'                     rc=0
  with slash                      -> intacta
  with dolar                      -> intacta
  with quotes                   -> intacta
  with space                    -> intacta
  with almohadilla                -> intacta
```

- `&` becomes the matched line, so `abc&def` is stored as `abcpassword: wazuh-wuidef`.
- `|` breaks the expression. `sed` exits 1 and writes nothing, so the file keeps the shipped default `wazuh-wui`. The error goes to stderr, and the entrypoint calls `/wazuh_dashboard_config.sh` without checking its result, so it scrolls past in the container log while the dashboard starts with the default password.
- `\` is consumed, so `abc\def` is stored as `abcdef`, with exit code 0 and no output at all.

## The dashboard indexer password

The same defect in the general `CONFIG_MAP` loop, which handles `opensearch.password` among 30 other settings:

```
docker run --rm -e OPENSEARCH_PASSWORD='abc&def' ... -c '/wazuh_dashboard_config.sh; grep ^opensearch.password ...'

opensearch.password: abc#opensearch.password:def
```

Here the `&` expanded to the commented placeholder line the substitution matched, so the stored password is `abc#opensearch.password:def`.

## The manager cluster key

`/etc/cont-init.d/0-wazuh-init` injects five cluster settings the same way:

```bash
sed -i "/<cluster>/,/<\/cluster>/ s|<node_name>.*</node_name>|<node_name>$WAZUH_NODE_NAME</node_name>|" "$WAZUH_MANAGER_CONF"
sed -i "/<cluster>/,/<\/cluster>/ s|<node_type>.*</node_type>|<node_type>$WAZUH_NODE_TYPE</node_type>|" "$WAZUH_MANAGER_CONF"
sed -i "/<cluster>/,/<\/cluster>/ s|<key>.*</key>|<key>$WAZUH_CLUSTER_KEY</key>|" "$WAZUH_MANAGER_CONF"
sed -i "/<cluster>/,/<\/cluster>/ s|<bind_addr>.*</bind_addr>|<bind_addr>$WAZUH_CLUSTER_BIND_ADDR</bind_addr>|" "$WAZUH_MANAGER_CONF"
```

Running the full init script in the image with each key and reading back the `<key>` element:

```
   sent 'key-normal-1234'  -> written 'key-normal-1234'                        intacta
   sent 'abc&def'            -> written '09184297c04d75e805538534681511a7</key>def' CORROMPIDA
   sent 'abc|def'            -> written '09184297c04d75e805538534681511a7'          CORROMPIDA
   sent 'abc\def'            -> written 'abcdef'                                    CORROMPIDA
   sent 'abc/def'            -> written 'abc/def'                                   intacta
```

The `&` case is the worst of the set: the expansion inserts the whole matched element, `<key>09184297c04d75e805538534681511a7</key>`, inside the new one, so the file ends up with nested `<key>` tags. The `|` case leaves the image's default key untouched, which means an operator who set a cluster key containing a pipe is running on the published default and has no indication of it (see [263](already-reported/BUG_263-the-cluster-key-that-authenticates-master-worker-communication-is-a-constant-in-the-public-image.md)).

## Why the characters matter

A `|` or a `&` in a generated password is not exotic. `openssl rand -base64` produces `+`, `/` and `=`, which are safe here, but every common password manager and most `pwgen`-style generators emit the full printable set, and `&` and `|` are in it. The values that pass through these substitutions are exactly the ones an operator is told to change from the default: the Wazuh API password, the indexer password and the cluster key.

## Reproducing

```
docker run --rm -e API_PASSWORD='abc&def' --entrypoint sh wazuh/wazuh-dashboard:5.0.0-latest \
  -c '/wazuh_dashboard_config.sh; grep -A6 "^wazuh_core.hosts:" /usr/share/wazuh-dashboard/config/opensearch_dashboards.yml'

docker run --rm -e WAZUH_CLUSTER_KEY='abc&def' -e WAZUH_NODE_NAME=n1 --entrypoint bash \
  wazuh/wazuh-manager:5.0.0-latest \
  -c 'bash /etc/cont-init.d/0-wazuh-init >/dev/null 2>&1; sed -n "/<cluster>/,/<\/cluster>/p" /var/wazuh-manager/etc/wazuh-manager.conf'
```

## Suggested fix

Escape the replacement text, not just the pattern. The scripts already escape the key for the left-hand side; the same treatment is needed on the right:

```bash
escape_repl() { printf '%s' "$1" | sed -e 's/[\\&|]/\\&/g'; }
...
        sed -i "s|^[#[:space:]]*${escaped_key}:.*|${key}: $(escape_repl "$value")|" "$FILE"
```

and the exit status of each `sed` should be checked, so a substitution that fails stops the container rather than leaving a default credential in place. For the manager, the same helper applied to `WAZUH_CLUSTER_KEY`, `WAZUH_NODE_NAME` and `WAZUH_CLUSTER_BIND_ADDR`.

A stronger fix for the manager is to stop editing XML with `sed` at all; the same script already builds the `<hosts>` and `<nodes>` blocks in temporary files and splices them in, which is where [261](BUG_261-five-of-the-eight-plausible-values-for-wazuh-indexer-hosts-produce-a-corrupt-host-and-the-init-reports-success.md) comes from.

## Environment

- Manager 5.0.0 (`wazuh/wazuh-manager:5.0.0-latest`) and dashboard 5.0.0 (`wazuh/wazuh-dashboard:5.0.0-latest`).
- `wazuh-docker` 5.0.0 branch at `37452c5c` (2026-08-24): `build-docker-images/wazuh-manager/config/0-wazuh-init` and `build-docker-images/wazuh-dashboard/config/wazuh_dashboard_config.sh`.
- Reproduced against the images alone; no running deployment needed.

## Appendix: a second path, in the same entrypoint, that corrupts a different password

The `sed` substitutions this finding is about are not the only place a credential is passed through
the shell unprotected. Four lines earlier in the same `/entrypoint.sh`, the dashboard's indexer
credentials go into the keystore through an unquoted expansion:

```sh
DASHBOARD_USERNAME="${DASHBOARD_USERNAME:-kibanaserver}"
DASHBOARD_PASSWORD="${DASHBOARD_PASSWORD:-kibanaserver}"
...
echo $DASHBOARD_USERNAME | ".../opensearch-dashboards-keystore" add opensearch.username --stdin --allow-root -f
echo $DASHBOARD_PASSWORD | ".../opensearch-dashboards-keystore" add opensearch.password --stdin --allow-root -f
```

Both are assigned with quotes and then expanded without them, so the shell word-splits and globs the
value on its way to `echo`. Measured inside the image, showing exactly what reaches the keystore's
standard input:

```
   password set by the operator     what the keystore receives
   correcta123                      correcta123
   con espacio aqui                 con espacio aqui
   dos  espacios                    dos espacios          <- two spaces became one
   tab<TAB>here                     tab here              <- the tab became a space
   glob*aqui                        glob*aqui             <- survived, see below
   signo?interrogacion              signo?interrogacion
   corchetes[ab]c                   corchetes[ab]c
```

Two definite corruptions: a run of whitespace collapses to a single space, and a tab becomes a
space. Either one means the dashboard authenticates to the indexer with a password the operator did
not set, and the only symptom is an authentication failure with no indication that the value was
rewritten between the compose file and the keystore.

The three glob cases came through unchanged here, and that is not a guarantee. `*`, `?` and `[...]`
are expanded against the working directory, which is `/usr/share/wazuh-dashboard` for this
entrypoint; they survived because nothing there matched. A password containing `*` in a directory
where something does match would be replaced by the matching filenames. The correctness of the value
should not depend on what files happen to sit beside the script.

Quoting both expansions is the whole fix, and it is the same one-character-per-line change as the
`sed` escaping this finding asks for:

```sh
   printf '%s\n' "$DASHBOARD_PASSWORD" | ... add opensearch.password --stdin --allow-root -f
```

`printf` rather than `echo` also removes the second hazard in that line: `echo` interprets
backslash escapes in some shells, so a password containing `\n` or `\t` is a third corruption
waiting on the shell the image happens to use.

## Appendix: the agent image does it too, with a different delimiter and a worse outcome

This finding says "both the manager image and the dashboard image". It is three images. The agent's
`/etc/cont-init.d/0-wazuh-init` writes three environment values into `ossec.conf` the same way, with
`#` as the delimiter instead of `|`:

```sh
  sed -i "s#<address>CHANGE_MANAGER_IP</address>#<address>$WAZUH_MANAGER_SERVER</address>#g" ${WAZUH_INSTALL_PATH}/etc/ossec.conf
  sed -i "s#<manager_address>CHANGE_ENROLL_IP</manager_address>#<manager_address>$WAZUH_REGISTRATION_SERVER</manager_address>#g" ${WAZUH_INSTALL_PATH}/etc/ossec.conf
  sed -i "s#<agent_name>CHANGE_AGENT_NAME</agent_name>#<agent_name>$WAZUH_AGENT_NAME</agent_name>#g" ${WAZUH_INSTALL_PATH}/etc/ossec.conf
```

The delimiter change moves which character breaks the expression, from `|` to `#`, and leaves `&`
and `\` dangerous as before. Measured by starting the published agent image with each value and
reading line 12 of the resulting `ossec.conf`:

```
   WAZUH_AGENT_NAME     resulting line

   agente-01            <agent_name>agente-01</agent_name>
   ag&ente              <agent_name>ag<agent_name>CHANGE_AGENT_NAME</agent_name>ente</agent_name>
   ag#ente              <agent_name>CHANGE_AGENT_NAME</agent_name>
   ag\ente              <agent_name>agente</agent_name>
```

Three different failures from three characters:

- **`&` produces malformed XML.** It expands to the whole matched text, so the replacement nests one
  `<agent_name>` element inside another and leaves an unbalanced pair. This is worse than the cases
  above, where the outcome is a wrong value in a well-formed file: here the configuration file the
  agent parses is no longer valid XML.
- **`#` leaves the placeholder.** It is the delimiter, so `sed` fails and writes nothing, and the
  agent starts with the literal string `CHANGE_AGENT_NAME` as its name. Nothing checks the exit
  status, so the only trace is one line on stderr among the startup output.
- **`\` is consumed.** The name silently becomes `agente`, one character shorter than the operator
  set, which is the quiet failure: the agent enrols successfully under a name nobody chose.

The same three characters apply to `WAZUH_MANAGER_SERVER` and `WAZUH_REGISTRATION_SERVER` on the two
lines above, where the value is an address rather than a name. An address is less likely to carry
them, but a hostname may, and the failure there is an agent pointed at the wrong manager or at the
literal `CHANGE_MANAGER_IP`.

The fix is the same escaping this finding already asks for, applied to a third script, and the `#`
delimiter needs `#` added to the escaped set rather than `|`.

jesusph04:
## Description 

The manager, dashboard and agent images inject configuration values (passwords, cluster keys, endpoints, names) with
`sed -i "s|pattern|literal ${value}|"`, without escaping the value. `&`, `\`, and the delimiter itself (`|` for manager/dashboard, `#` for agent) are all special on the replacement side of a sed substitution, and none was handled — silently corrupting or discarding credentials the operator explicitly set, always with exit 0.

A separate defect in the same dashboard entrypoint piped `$DASHBOARD_USERNAME`/`$DASHBOARD_PASSWORD` into the keystore via unquoted `echo`, collapsing whitespace runs and turning tabs into spaces.

## Fix

- Added `escape_repl()` (`printf '%s' "$1" | sed -e 's/[\\&|]/\\&/g'`)   to the manager and dashboard scripts, applied to every value on the  right-hand side of a `sed` substitution.
- Same helper on the agent script, with `#` in the escaped set instead of `|`, matching its delimiter.
- Quoted the two `echo $VAR | keystore add --stdin` lines in the  dashboard entrypoint and switched to `printf '%s\n'`.

## Validation

Reproduced every scenario from the issue against a bind-mounted copy of each patched script over the published `5.0.0-latest` images (no rebuild needed — bug is purely in shell scripts).

**Dashboard (`API_PASSWORD`, `wazuh_core.hosts` block):**

```
abc&def -> password: "abc&def" rc=0
abc|def -> password: "abc|def" rc=0
abc\def -> password: "abc\def" rc=0
```

**Manager (`WAZUH_CLUSTER_KEY`):**

```
abc&def -> <key>abc&def</key> rc=0
abc|def -> <key>abc|def</key> rc=0
abc\def -> <key>abc\def</key> rc=0
```

**Agent (`WAZUH_AGENT_NAME`):**

```
agente&01 -> <agent_name>agente&01</agent_name> rc=0
agente#01 -> <agent_name>agente#01</agent_name> rc=0
agente\01 -> <agent_name>agente\01</agent_name> rc=0
```

**Dashboard entrypoint (keystore quoting), isolated before/after:**

```
  before (echo, unquoted)   after (printf, quoted)

'dos espacios' -> 'dos espacios' -> 'dos espacios'
'tab\there' -> 'tab here' -> 'tab\there'
```

All previously-corrupting inputs (`&`, `|`/`#`, `\`, whitespace runs,
tabs) now pass through unchanged, with `rc=0` in every case.


jotamolas:
# Review

Good, comprehensive fix — covers all 3 scripts (agent, dashboard, manager) with a consistent `escape_repl()` per delimiter. CI is green.

One residual gap, verified: for the dashboard API `password` field specifically, this escapes for `sed`'s own special characters (`&`, `\`, `|`) but writes the result inside **double quotes** (`password: \"$(escape_repl "$API_PASSWORD")\"`). Double-quoted YAML scalars have their own escaping rules, and a password containing a literal `"` is not covered by `escape_repl` — it breaks the YAML outright.

Confirmed with a quick reproduction (`abc"def` as the password): writing `password: "abc"def"` fails to parse as YAML at all (`expected <block end>, but found <scalar>`). That's arguably worse than the original bug — instead of a silently wrong value, the dashboard config becomes invalid.

There's a community PR open on the same issue, [#2645](https://github.com/wazuh/wazuh-docker/pull/2645) (dashboard-only, by an external contributor), that sidesteps this by switching the API password to a **single-quoted** YAML scalar instead, with YAML's own escaping (doubling any embedded `'`):

```bash
yaml_api_password=$(printf '%s' "$API_PASSWORD" | sed "s/'/''/g")
escaped_api_password=$(escape_repl "$yaml_api_password")
# then: s|password:.*|password: '$escaped_api_password'|
```

I verified that approach round-trips `abc"def` correctly. Worth adopting the same single-quote treatment here for the `password` line specifically (the other dashboard fields, and the manager/agent XML fields, don't have this YAML-quoting concern, so `escape_repl` alone is fine there).

patriiciacoboss (approved):

PR #2647, opened 2026-09-08
Validate and default WAZUH_INDEXER_HOSTS parsing

### LINKED ISSUE #2630: Five of the eight plausible values for WAZUH_INDEXER_HOSTS produce a corrupt host and the init reports success
## Summary

The manager image builds the `<indexer><hosts>` block from `WAZUH_INDEXER_HOSTS` by splitting each entry on a colon with shell parameter expansion, without checking that a colon is there or that what follows it is a port. Omitting the port produces `https://wazuh.indexer:wazuh.indexer`. Including a scheme, which the variable's name and the value's URL-like shape both invite, produces `https://https://wazuh.indexer://wazuh.indexer:9200`. Both IPv6 forms are mangled, and an entry after a comma keeps its leading space. In every case the init script prints `Wazuh manager config modified successfully.` and exits 0.

## The code

`/etc/cont-init.d/0-wazuh-init`, lines 265 to 280:

```bash
if [[ -n "$WAZUH_INDEXER_HOSTS" ]]; then
    TMP_HOSTS=$(mktemp)
    {
        echo "    <hosts>"
        IFS=',' read -ra NODES <<< "$WAZUH_INDEXER_HOSTS"
        for NODE in "${NODES[@]}"; do
            IP="${NODE%:*}"
            PORT="${NODE#*:}"
            echo "      <host>https://$IP:$PORT</host>"
        done
        echo "    </hosts>"
    } > "$TMP_HOSTS";
    sed -i -e '/<indexer>/,/<\/indexer>/{ /<hosts>/,/<\/hosts>/{ /<hosts>/r '"$TMP_HOSTS" \
            -e 'd }}' "$WAZUH_MANAGER_CONF";
    rm -f "$TMP_HOSTS";
fi
```

`${NODE%:*}` and `${NODE#*:}` both return the whole string when there is no colon, so a bare hostname is emitted as its own port. `${NODE#*:}` cuts at the *first* colon, so a scheme or a bracketed IPv6 address is split in the wrong place.

## Measured

Running that exact fragment against the shipped `wazuh-manager.conf` inside the image, and printing the resulting `<host>` lines:

```
  wazuh.indexer:9200 (lo that uses the compose) -> <host>https://wazuh.indexer:9200</host>
  wazuh.indexer (without port)                 -> <host>https://wazuh.indexer:wazuh.indexer</host>
  https://wazuh.indexer:9200 (with schema)   -> <host>https://https://wazuh.indexer://wazuh.indexer:9200</host>
  http://wazuh.indexer:9200                  -> <host>https://http://wazuh.indexer://wazuh.indexer:9200</host>
  n1:9200,n2:9200,n3:9200 (three nodos)       -> <host>https://n1:9200</host> <host>https://n2:9200</host> <host>https://n3:9200</host>
  n1:9200, n2:9200 (with space)             -> <host>https://n1:9200</host> <host>https:// n2:9200</host>
  [::1]:9200 (IPv6 in brackets)          -> <host>https://[::1]::1]:9200</host>
  ::1:9200 (IPv6 without brackets)              -> <host>https://::1::1:9200</host>
```

Three of the eight are right: the exact `host:port` form the compose happens to use, and a comma-separated list of that form with no spaces. Everything else is silently corrupt.

Running the full init script in the image with the port omitted confirms it end to end:

```
docker run --rm -e WAZUH_INDEXER_HOSTS=wazuh.indexer -e WAZUH_NODE_NAME=test \
  --entrypoint bash wazuh/wazuh-manager:5.0.0-latest -c 'bash /etc/cont-init.d/0-wazuh-init; ...'
```

```
   Wazuh manager config modified successfully.
   Creating the manager server certificate and key (/var/wazuh-manager/etc/certs/remoted.pem)
   === codigo: 0
   === block indexer resulting:
     <indexer>
       <hosts>
         <host>https://wazuh.indexer:wazuh.indexer</host>
       </hosts>
       ...
```

The script reports success. Nothing downstream re-checks the value, so the first sign of trouble is the manager failing to reach the indexer at run time, with a host string that a reader has to look at twice to see is wrong.

## Why these forms are plausible

The variable is named `WAZUH_INDEXER_HOSTS`, plural, and its value looks like a URL without a scheme. The three wrong forms above are what an operator writes when copying from adjacent places in the same product:

- the manager's own `wazuh-manager.conf` writes the value with a scheme, `<host>https://127.0.0.1:9200</host>`, so `https://wazuh.indexer:9200` is the natural thing to paste;
- the dashboard service in the same compose file takes `OPENSEARCH_HOSTS=https://wazuh.indexer:9200`, with the scheme, so the two sibling services in one file want different formats for the same address;
- the indexer's own default port is well known, so leaving it off is an easy assumption.

Nothing in the compose file, the Dockerfile or the script states the required format.

## Suggested fix

Validate and default, rather than splitting blind:

```bash
        for NODE in "${NODES[@]}"; do
            NODE="${NODE#"${NODE%%[![:space:]]*}"}"      # trim leading space
            NODE="${NODE#http://}"; NODE="${NODE#https://}"
            if [[ "$NODE" == *:* ]]; then
                HOST="${NODE%:*}"; PORT="${NODE##*:}"
            else
                HOST="$NODE"; PORT=9200
            fi
            [[ "$PORT" =~ ^[0-9]+$ ]] || { print "Invalid indexer host '$NODE'"; exit 1; }
            echo "      <host>https://${HOST}:${PORT}</host>"
        done
```

`${NODE##*:}` instead of `${NODE#*:}` also fixes the bracketed IPv6 case. Whatever the chosen shape, the script should refuse a value it cannot parse rather than writing it into the configuration and reporting success.

## Environment

- Manager 5.0.0, image `wazuh/wazuh-manager:5.0.0-latest`.
- `wazuh-docker` 5.0.0 branch at `37452c5c` (2026-08-24), `build-docker-images/wazuh-manager/config/0-wazuh-init` and both compose files.
- Reproduced against the image alone; no running deployment needed.

## The other list variable in the same script uses the opposite separator

`WAZUH_CLUSTER_NODES` is built a few lines below `WAZUH_INDEXER_HOSTS` in the same file, and it
splits on whitespace rather than on commas:

```bash
if [[ -n "$WAZUH_CLUSTER_NODES" ]]; then
    TMP_NODES=$(mktemp)
    {
        echo "    <nodes>"
        for N in $WAZUH_CLUSTER_NODES; do
            echo "        <node>$N</node>"
        done
        echo "    </nodes>"
    } > "$TMP_NODES";
```

against `IFS=',' read -ra NODES <<< "$WAZUH_INDEXER_HOSTS"` for the indexer hosts. Running that
fragment against the shipped configuration:

```
  a only node (lo that uses the compose)           -> <node>wazuh.master</node>
  two nodos separados by space                -> <node>wazuh.master</node> <node>wazuh.worker</node>
  two nodos separated by comma                   -> <node>wazuh.master,wazuh.worker</node>
  two nodos, coma and space                      -> <node>wazuh.master,</node> <node>wazuh.worker</node>
  three nodos by coma                            -> <node>n1,n2,n3</node>
```

Both shipped compose files pass a single value, `WAZUH_CLUSTER_NODES=wazuh.master`, so the shipped
deployments never exercise the separator. An operator adding a second manager node has two list
variables in front of them, set in the same `environment:` block, and the one they are copying from
is comma-separated:

```yaml
      - WAZUH_INDEXER_HOSTS=wazuh1.indexer:9200,wazuh2.indexer:9200,wazuh3.indexer:9200
      - WAZUH_CLUSTER_NODES=wazuh.master
```

Writing `WAZUH_CLUSTER_NODES=wazuh.master,wazuh.worker` by analogy produces one node entry whose name
is the whole string, and the init prints `Wazuh manager config modified successfully.` exactly as it
does for the indexer host case. Neither variable is documented anywhere, so there is nothing to
consult; see
[265](BUG_265-the-deployment-documentation-attributes-five-environment-variables-to-the-wrong-container-and-describes-an-override-rule-that-works-for-neither-file.md).

patriiciacoboss:
## Changes
Replaced the blind split with validate-and-default logic:

```bash
NODE="${NODE#"${NODE%%[![:space:]]*}"}"      # trim leading space
NODE="${NODE#http://}"; NODE="${NODE#https://}"
if [[ "$NODE" == *:* ]]; then
    HOST="${NODE%:*}"; PORT="${NODE##*:}"    # last colon, not first — fixes bracketed IPv6
else
    HOST="$NODE"; PORT=9200
fi
[[ "$PORT" =~ ^[0-9]+$ ]] || { print "Invalid indexer host '$NODE'" >&2; exit 1; }
```

An unparseable host now fails loudly (non-zero exit, clear message) instead of silently writing a broken config and reporting success.

**Extra fix found while testing:** the error message was being swallowed. The `for` loop lives inside a `{ ... } > "$TMP_HOSTS"` block, so `print`'s stdout (including the error message) was captured into the temp file instead of reaching the log — the script did exit non-zero, but "Invalid indexer host" never actually showed up anywhere. Redirected it to stderr (`>&2`) so it's visible.

## Verification

Tested all 9 cases from the issue against the real script, inside a `5.0.0` image built locally from real packages 

| Input | Result |
|---|---|
| `wazuh.indexer:9200` | `https://wazuh.indexer:9200` ✅ |
| `wazuh.indexer` (no port) | `https://wazuh.indexer:9200` ✅ |
| `https://wazuh.indexer:9200` | `https://wazuh.indexer:9200` ✅ |
| `http://wazuh.indexer:9200` | `https://wazuh.indexer:9200` ✅ |
| `n1:9200,n2:9200,n3:9200` | 3 separate `<host>` lines ✅ |
| `n1:9200, n2:9200` (space) | 2 lines, no leading space ✅ |
| `[::1]:9200` (IPv6, brackets) | `https://[::1]:9200` ✅ |
| `::1:9200` (IPv6, no brackets) | `https://::1:9200` (parsed sensibly) ✅ |
| unparseable garbage | exit 1, `Invalid indexer host '...'` printed, config left untouched ✅ |

Each case was run like this, against the real `0-wazuh-init` script inside the built image:

```bash
docker run --rm --entrypoint bash -e WAZUH_INDEXER_HOSTS="<value>" wazuh/wazuh-manager:test-indexer-parsing -c "
mkdir -p /var/wazuh-manager/etc
printf '<ossec_config>\n  <indexer>\n    <hosts>\n    </hosts>\n  </indexer>\n</ossec_config>\n' > /var/wazuh-manager/etc/wazuh-manager.conf
bash /etc/cont-init.d/0-wazuh-init 2>&1
cat /var/wazuh-manager/etc/wazuh-manager.conf
"
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,7 @@
 
 | Issue | Comment |
 | - | - |
+| [#2630](https://github.com/wazuh/wazuh-docker/issues/2630) | Fixed WAZUH_INDEXER_HOSTS parsing: hosts without an explicit port, URLs with a scheme, and bracketed IPv6 addresses were silently mangled instead of being parsed correctly, and unparseable values were written to the config without any error |
 | [#2634](https://github.com/wazuh/wazuh-docker/issues/2634) | Default WAZUH_CLUSTER_BIND_ADDR to 0.0.0.0 so the published manager image actually starts |
 | [#2604](https://github.com/wazuh/wazuh-docker/issues/2604) | Adapt Wazuh manager image to new enroll process |
 | [#2594](https://github.com/wazuh/wazuh-docker/issues/2594) | Added `diffutils` to the Wazuh agent image so FIM `report_changes` can generate content diffs |
--- a/build-docker-images/wazuh-manager/config/etc/cont-init.d/0-wazuh-init
+++ b/build-docker-images/wazuh-manager/config/etc/cont-init.d/0-wazuh-init
@@ -269,9 +269,15 @@ if [[ -n "$WAZUH_INDEXER_HOSTS" ]]; then
         echo "    <hosts>"
         IFS=',' read -ra NODES <<< "$WAZUH_INDEXER_HOSTS"
         for NODE in "${NODES[@]}"; do
-            IP="${NODE%:*}"
-            PORT="${NODE#*:}"
-            echo "      <host>https://$IP:$PORT</host>"
+            NODE="${NODE#"${NODE%%[![:space:]]*}"}"
+            NODE="${NODE#http://}"; NODE="${NODE#https://}"
+            if [[ "$NODE" == *:* ]]; then
+                HOST="${NODE%:*}"; PORT="${NODE##*:}"
+            else
+                HOST="$NODE"; PORT=9200
+            fi
+            [[ "$PORT" =~ ^[0-9]+$ ]] || { print "Invalid indexer host '$NODE'" >&2; exit 1; }
+            echo "      <host>https://${HOST}:${PORT}</host>"
         done
         echo "    </hosts>"
     } > "$TMP_HOSTS";


vcerenu (approved):
LGTM

Issue #1578, opened 2024-10-16
File-owner and/or -permissions in config-folder won't be adjusted at first run (Bug?)


n1smithy:
Hi @ all,

this is my first time I'm using wazuh, and I hope my containers run stablely now!

However: As a security-feature docker offers a userspace-remapping so docker-containers run outside of the host's userspace. But there's a little issue when using this feature.

The shared/persistent storaged folders/files must be chowned by root to fit in that userspace. Usually one sets all of the persistent files to the first UID and GID of the designated remapping-userspace. For example: the remap-userspace begins with UID/GID 100000:100000, then this would be the UID and GID for the remapped root-superuser. When running the container at it's first time, the shared files and folders should be automatically be set to them's right ownerships by the containers (remapped) root-user.

However, within wazuh's container-instances it's impossible to let the server write to the .../config/wazuh-dashboard/wazuh.yml-file because of wrong file-permissions/-ownership.

If there's a list of the right ownerships and permissions, one could at least easily change them manually.

Thanks for reading!

Greetings!

PR #2651, opened 2026-09-09
Supervise manager daemons from entrypoint and exit on failure

### LINKED ISSUE #2626: Nothing restarts the manager API when its master process dies, and the shipped container detects the failure and then does nothing about it
Kill the wazuh-manager API's master process and the API never comes back. No supervisor respawns it. The manager's own control script notices within seconds and reports `wazuh-manager-apid not running...`, the container health check notices after about two minutes and turns the container `unhealthy`, and neither observation causes anything to happen: `RestartCount` stays at 0, the container stays `running`, and the API stays dead until somebody restarts the container by hand.

Every other manager daemon keeps running throughout. Agents stay connected, events keep flowing, alerts keep being written. The only thing that is gone is the management plane, which is also the thing an operator would use to find out that anything is wrong.

|Wazuh version|Component|Install type|Install method|Platform|
|---|---|---|---|---|
| 5.0.0-latest | wazuh-manager API (`api/scripts/wazuh_manager_apid.py`), process supervision, and the container health check shipped with the image | Manager | Docker (public ECR wazuh-cicd) | Amazon Linux 2023 (agents), two-node cluster |

**Severity:** High. The outage is permanent and unattended, the deployment's own health check proves the failure is detectable, and [070](not-for-filing/duplicated/DUPLICATED_070-rejected-request-bodies-set-the-api-memory-footprint.md) provides a way for an unauthenticated client to cause it in two seconds.

## The process tree

The API runs as a master with three children:

```
   pid=160     ppid=1      python3 .../api/scripts/wazuh_manager_apid.py
   pid=161     ppid=160    python3 .../api/scripts/wazuh_manager_apid.py
   pid=162     ppid=160    python3 .../api/scripts/wazuh_manager_apid.py
   pid=165     ppid=160    python3 .../api/scripts/wazuh_manager_apid.py
```

Pid 160 owns the listening socket. The three children are re-parented to init when it dies, and they stay alive, doing nothing, for as long as the container runs.

## What happens when the master dies

`kill -9` on pid 160, with no memory pressure and nothing else touched:

```
   seconds   API      health        reinicios   corriendo
   0s      000      healthy      0           true
   20s     000      healthy      0           true
   40s     000      healthy      0           true
   60s     000      healthy      0           true
   80s     000      healthy      0           true
   100s    000      healthy      0           true
   120s    000      unhealthy    0           true
   140s    000      unhealthy    0           true
   ...
   240s    000      unhealthy    0           true

   health=unhealthy fallos_seguidos=10 reinicios=0 corriendo=true
```

`000` is curl reporting that no HTTP response arrived at all. It is `000` at every sample from the moment the master dies until the container is restarted by hand, at which point the API returns in about ten seconds. There is no respawn, no back-off, no retry: the API is simply gone.

## Four ways of monitoring this, and what each one says

While the API is dead:

```
   --- the truth: a request HTTP real
   API responde: 000

   --- 1. Wazuh's own control script
      wazuh-manager-clusterd is running...
      wazuh-manager-modulesd is running...
      wazuh-manager-monitord is running...
      wazuh-manager-remoted is running...
      wazuh-manager-analysisd is running...
      wazuh-manager-db is running...
      wazuh-manager-authd is running...
      wazuh-manager-apid not running...

   --- 2. API process count (what a 'ps' check looks at)
      procesos apid live: 3

   --- 3. container state
      corriendo=true health=unhealthy reinicios=0

   --- 4. TCP port check
      conexion al 55000 accepts: si
      listener someone inside of the container: NO
```

Only the first is both correct and immediate. `wazuh-manager-control status` gets it right within seconds, which is the proof that the failure is trivially detectable from inside the container. Nothing consumes that answer.

The other three are each wrong in a different way, and each corresponds to a monitoring style someone actually uses:

- **Counting processes** sees three live `apid` processes and concludes the API is fine. The orphans keep a `pgrep`-style check green forever.
- **The container state** says `running`, because PID 1 of the container is the manager's entrypoint, not the API. An orchestrator watching liveness by container state sees nothing.
- **A TCP check on 55000** connects successfully, because Docker's port proxy on the host accepts the connection whether or not anything listens inside. Reading `/proc/net/tcp` in the container shows the listening ports are `[1514, 1515, 1516, 1517, 36029]`: 55000 is not among them. A load balancer doing a connect check keeps the node in rotation and every request it forwards hangs.

## The health check is correct and is ignored

The image does ship a health check, and it is written correctly:

```json
{
    "Test": ["CMD-SHELL",
        "/var/wazuh-manager/bin/wazuh-manager-control status 2>/dev/null | grep -q 'not running' && exit 1 || exit 0"],
    "Interval": 15000000000,
    "Timeout": 5000000000,
    "StartPeriod": 60000000000,
    "Retries": 5
}
```

It consults the one surface that tells the truth, and after five consecutive failures at fifteen-second intervals it correctly flips the container to `unhealthy`, which is what the table above shows happening at around 120 seconds.

And then nothing. The container's restart policy is `always`, and Docker restart policies do not react to health status: they act on the container's main process exiting, which never happens here. So the deployment is configured to detect exactly this failure and has no path from detecting it to acting on it. The container sits `unhealthy` with `FailingStreak=10` and climbing, `RestartCount=0`, indefinitely.

## Why this is not only a manual-kill scenario

The kill above was done by hand to isolate supervision from the cause of death, and to show that the permanence has nothing to do with *how* the master died. But the master does not need to be killed by hand.

[070](not-for-filing/duplicated/DUPLICATED_070-rejected-request-bodies-set-the-api-memory-footprint.md) documents that every request body is buffered and deserialised by the access-logging middleware before authentication, so an unauthenticated client sets the API process's memory high-water mark. Ten unauthenticated 9 MB bodies to `POST /agents`, sent in two seconds and all answered `401`, are enough to have the kernel kill this master under any container memory limit. The appendix to 070 records that measurement. The two findings compose: 070 supplies a remote, credential-free trigger, and this one supplies the permanence.

## Expected behavior

A dead API master is respawned, with back-off, by whatever supervises the manager's daemons, exactly as the other seven daemons in `wazuh-manager-control status` would be. Failing that, the orphaned children are reaped rather than left to keep process-count checks green, and the health check's verdict is wired to something that acts on it.

## Steps to reproduce

1. Confirm the API answers: `curl -sk -o /dev/null -w '%{http_code}' -X POST -u wazuh:wazuh 'https://localhost:55000/security/user/authenticate?raw=true'` returns `200`.
2. Find the master, the `apid` process whose parent is pid 1: `docker exec <manager> sh -c "ps -eo pid,ppid,args | grep '[a]pid' | awk '\$2==1{print \$1}'"`.
3. `docker exec <manager> kill -9 <pid>`.
4. Repeat step 1 every fifteen seconds for four minutes. It returns `000` every time.
5. `docker inspect <manager> --format '{{.State.Health.Status}} {{.RestartCount}} {{.State.Running}}'` returns `unhealthy 0 true` from about two minutes onward, and stays there.
6. `docker restart <manager>` brings the API back in about ten seconds.

Steps 3 to 6 were run three times with the same result, and once more with the death caused by memory exhaustion rather than a signal, also with the same result.

jotamolas:
## Summary

`wazuh-manager-control` starts all 8 manager daemons once and the entrypoint's `tail -F & wait` never checks whether any of them are still alive. If one dies (e.g. `apid`'s master killed by OOM or a signal, see #2626), the container stays `Up` forever with a dead API, and `restart: always` never gets a chance to react because nothing ever exits.

## Changes

Added a supervision loop to the manager's `entrypoint.sh`: every 15s (after a 60s grace period matching the healthcheck's own `start_period`), it checks `wazuh-manager-control status`'s own exit code — same failure vocabulary as the compose healthcheck fixed in #2642 (`not running` / `failed to start` / `refused its configuration`) — and exits non-zero if a critical daemon is down, so the container actually stops and the existing `restart: always` policy brings it back.

Two exclusions were needed because `wazuh-manager-control`'s `status()` doesn't apply them itself:
- `apid` on worker nodes (it never runs there — same exclusion the multi-node worker healthcheck already applies).
- `authd` when `auth.disabled: true` is set (`start_service()` skips starting it in that case, but `status()` has no matching exclusion — a real bug in the core script found while testing this fix; without compensating here, any deployment with `auth.disabled: true` would restart-loop forever).

## Verification

Reproduced live against the published image (`public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest`), mounting the modified entrypoint over it:

- **Baseline repro**: killed the `apid` master process directly (`kill -9` on the pid with `ppid=1`). Confirmed the container previously stayed `Up` indefinitely; with this fix, it exits ~60-70s later with `"Wazuh Manager is not running as expected, exiting so the container can be restarted..."`, and with `--restart=always`, `RestartCount` climbs 0→1→2→3 in ~70s cycles instead of staying at 0.
- **`auth.disabled: true` case**: set `<disabled>yes</disabled>` in `wazuh-manager.conf`, confirmed `authd` correctly doesn't start but `wazuh-manager-control status` still reports it `not running` (the core bug). Confirmed the exclusion neutralizes this specific false positive.
- **Negative control**: with `auth.disabled: false`, killing `authd` for real is still correctly detected as unhealthy (the exclusion isn't masking real failures).
- `shellcheck` (style level) on the modified script: clean, no findings.

## Related

- #2627 / #2642: same failure vocabulary, same file area. This loop is the piece that makes the (now correct) healthcheck's `unhealthy` status actually result in an action — under plain `docker compose`, nothing else consumes that status.
- Confirmed this isn't a Docker-vs-others gap: `wazuh-kubernetes` already has equivalent liveness probes (#1578) that do restart the pod on the same failure signal — this PR is Docker reaching parity with what Kubernetes already does, not a new workaround.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,7 @@
 
 | Issue | Comment |
 | - | - |
+| [#2626](https://github.com/wazuh/wazuh-docker/issues/2626) | Supervise the manager daemons from the entrypoint and exit if a critical one dies, so `restart: always` can bring the container back instead of it staying up indefinitely with a dead API |
 | [#2634](https://github.com/wazuh/wazuh-docker/issues/2634) | Default WAZUH_CLUSTER_BIND_ADDR to 0.0.0.0 so the published manager image actually starts |
 | [#2630](https://github.com/wazuh/wazuh-docker/issues/2630) | Fixed WAZUH_INDEXER_HOSTS parsing: hosts without an explicit port, URLs with a scheme, and bracketed IPv6 addresses were silently mangled instead of being parsed correctly, and unparseable values were written to the config without any error |
 | [#2627](https://github.com/wazuh/wazuh-docker/issues/2627) | Fixed manager and dashboard healthchecks to correctly detect failures instead of always reporting healthy |
--- a/build-docker-images/wazuh-manager/config/entrypoint.sh
+++ b/build-docker-images/wazuh-manager/config/entrypoint.sh
@@ -30,13 +30,15 @@ trap '_stop 0' SIGTERM SIGINT SIGQUIT
 STARTUP_GRACE_PERIOD=60
 CHECK_INTERVAL=15
 
-# Same default as cont-init.d/0-wazuh-init: master unless WAZUH_NODE_TYPE=worker
-# is set (only multi-node/docker-compose.yml's wazuh.worker service sets it).
-WAZUH_NODE_TYPE="${WAZUH_NODE_TYPE:-master}"
-
 # Same reader the core wazuh-manager-control script itself uses to decide
 # whether to start authd (start_service(), "auth.disabled: true" case).
 WAZUH_MANAGER_CONF="/var/wazuh-manager/bin/wazuh-manager-conf -H /var/wazuh-manager -f /var/wazuh-manager/etc/wazuh-manager.conf"
+
+# Timeout for each status check: wazuh-manager-control status has no upper
+# bound of its own, and a hang here would otherwise block both failure
+# detection and the SIGTERM trap below (docker stop would have to wait out
+# its own timeout and fall back to SIGKILL).
+STATUS_TIMEOUT=5
 
 # Supervise the manager: if a critical daemon is not running, exit non-zero so
 # the container dies and `restart: always` (single-node/multi-node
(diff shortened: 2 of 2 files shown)


jotamolas:
## QA pass found and fixed 2 reliability gaps in the original version of this PR

Ran 20 test cases against the initial commit (killing each of the 7 daemons individually, timing edge cases around the grace period, SIGTERM racing with failure detection, worker-node behavior, restart-loop overhead, etc.) — 18 passed, including a full E2E run against a real indexer with `restart: always` (confirmed `RestartCount` incrementing and daemons coming back healthy). Two real gaps surfaced and are fixed in the latest commit:

1. **Case-sensitive `apid` exclusion mismatch.** The original version excluded `apid` from the check based on the raw `WAZUH_NODE_TYPE` env var (`!= "master"`), while `0-wazuh-init` decides the real node type with the opposite check (`!= "worker"`). `WAZUH_NODE_TYPE=WORKER` (capitalized) resulted in a real master node (where `apid` runs) whose check nonetheless excluded `apid` — silently reproducing #2626 for that misconfiguration. Fixed by removing the manual exclusion entirely: `wazuh-manager-control status` already excludes `apid` correctly by reading the real `cluster.node_type` from the config file, so the duplicate check was both redundant and a source of bugs.

2. **Unbounded `status` call.** If `wazuh-manager-control status` itself hangs, the loop hung with it, and — worse — blocked the SIGTERM trap, so `docker stop` had to wait out its own timeout and fall back to SIGKILL. Fixed with `timeout 5` around the call, with the timeout case (exit 124) explicitly treated as unhealthy rather than letting empty output read as "no failures found".

Both reproduced live before and after the fix.

vcerenu (approved):
LGTM

Issue #2619, opened 2026-09-02
Provide the agent's manager CA in the Docker agent image (verification_mode now enforced by default)


jotamolas:
## Description

wazuh/wazuh#38684 reported that the agent's HTTPS transport performs no TLS verification by default. wazuh/wazuh#38786 fixes this on branch `5.0.0` (currently stage `rc1`): `verification_mode` moves from `none` to enforced (`system` trust store, or `certificate` when a CA is configured), and adds `WAZUH_REGISTRATION_CA` as the install-time variable that supplies the CA.

The Docker agent image's init script (`build-docker-images/wazuh-agent/config/etc/cont-init.d/0-wazuh-init`) writes `WAZUH_MANAGER_ENDPOINT`, `WAZUH_AGENT_NAME` and `WAZUH_REGISTRATION_PASSWORD` into `ossec.conf`, but has no equivalent for a CA. Our docker-compose stacks (`single-node`, `multi-node`) use self-signed certificates for the manager's HTTPS listener, so once the new default ships, the agent container will fail closed against a manager it previously connected to without any verification.

## Changes to be made

- Add an env var (mirror install-time `WAZUH_REGISTRATION_CA` from wazuh/wazuh#38786, e.g. `WAZUH_MANAGER_CA`) to `0-wazuh-init`, writing it into `<agent><ssl><certificate_authorities>`.
- Support mounting the CA file, the same way `WAZUH_CONFIG_MOUNT` already supports mounting a full `ossec.conf`.
- Update `single-node/docker-compose.yml` and `multi-node/docker-compose.yml` to pass the manager's own generated root CA to the agent service by default.
- Update the repo README for the new variable.

## Validations

- Fresh `docker compose up` (single-node and multi-node), no manual cert handling: agent enrolls and stays connected, no TLS verification warning in the agent log.
- A container started without a CA fails closed with a clear log message (matches upstream's own evidence in wazuh/wazuh#38786), instead of silently connecting unverified.

## Reference

- wazuh/wazuh#38684 (upstream report)
- wazuh/wazuh#38786 (upstream fix, targets `5.0.0`, stage `rc1`)
- wazuh-automation#3752 (related, closed — endpoint unification epic, same Objective "HTTPS comms")

vcerenu:
## Update

I have created a branch with the work done:
[change/2619-ca-verification](https://github.com/wazuh/wazuh-docker/tree/change/2619-ca-verification)

fcaffieri:
[...]

PR #2653, opened 2026-09-09
Fix agent re creation and re-enrollment

### LINKED ISSUE #2632: The shipped agent container registers a new agent on every recreation, and pinning its name makes it unable to re-enroll
## Summary

The agent compose declares no volumes, the agent image declares none and has no permanent-data mechanism, so `/var/ossec/etc/client.keys` is lost whenever the container is recreated. The default agent name is `wazuh-agent-$HOSTNAME`, and in Docker `$HOSTNAME` is the container id, so every recreation enrolls a brand new agent. The old registration stays on the manager forever, with its inventory and FIM documents, because nothing removes it and authd's id counter never reuses an id. Pinning `WAZUH_AGENT_NAME` to a stable value swaps that problem for a worse one: the shipped manager configuration has no `<force>` block, so the second enrollment under an existing name is refused as a duplicate and the agent never reconnects.

## No state is kept

`wazuh-agent/docker-compose.yml` on the 5.0.0 branch, in full:

```yaml
services:
  wazuh.agent:
    image: wazuh/wazuh-agent:5.0.0
    restart: always
    environment:
      - WAZUH_MANAGER_SERVER=<WAZUH_MANAGER_IP>
      - WAZUH_REGISTRATION_PASSWORD=<authd.pass-PASSWORD>
```

No `volumes:` key. The image agrees:

```
=== volumenes declarados by the image of the agent:
null
=== the image of the agent has permanent_data?
ls: cannot access '/permanent_data.env': No such file or directory
```

The manager image carries `/permanent_data.env` naming the five directories it restores from volumes; the agent image has no equivalent, so nothing under `/var/ossec` survives, `client.keys` included.

## A new identity per container

`/etc/cont-init.d/0-wazuh-init`, line 8:

```bash
WAZUH_AGENT_NAME=${WAZUH_AGENT_NAME:-"wazuh-agent-$HOSTNAME"}
```

Docker sets `$HOSTNAME` to the container's short id unless told otherwise, and it is a new id every time:

```
  hostname=22d0c77c67ed  ->  agent wazuh-agent-22d0c77c67ed
  hostname=23bcaf193ba6  ->  agent wazuh-agent-23bcaf193ba6
```

The compose sets only `WAZUH_MANAGER_SERVER` and `WAZUH_REGISTRATION_PASSWORD`, so this default is what a deployment gets. The enrollment block the init fills in has nothing that would reclaim a previous identity:

```xml
    <enrollment>
      <enabled>yes</enabled>
      <manager_address>CHANGE_ENROLL_IP</manager_address>
      <agent_name>CHANGE_AGENT_NAME</agent_name>
      <authorization_pass_path>etc/authd.pass</authorization_pass_path>
    </enrollment>
```

So each `docker compose down && up`, each image upgrade, each host reboot and each redeploy in an orchestrator produces one more agent on the manager. The previous one is never removed, and authd states plainly that its id is gone for good:

```
   Raising the agent id counter from %d to %d: ids that were handed out are never reused.
```

The abandoned registrations keep their documents too. Deletion of an agent's state documents only happens when the agent is explicitly removed, which queues the deletion in authd; an agent nobody removes keeps its rows in `wazuh-states-fim-files`, `wazuh-states-inventory-*` and `wazuh-states-sca` indefinitely, and those indices have no lifecycle policy.

## Pinning the name is worse

The obvious fix from an operator's side is to set `WAZUH_AGENT_NAME` to something stable. That makes the second enrollment a duplicate, and the shipped manager configuration does not allow duplicates to be replaced. The `<auth>` block in the manager image:

```xml
  <auth>
    <disabled>no</disabled>
    <port>1515</port>
    <use_source_ip>no</use_source_ip>
    <purge>yes</purge>
    <use_password>yes</use_password>
    <ciphers>TLS_AES_256_GCM_SHA384:TLS_CHACHA20_POLY1305_SHA256:TLS_AES_128_GCM_SHA256</ciphers>
    <!-- <ssl_agent_ca></ssl_agent_ca> -->
    <ssl_verify_host>no</ssl_verify_host>
    <ssl_manager_cert>etc/certs/remoted.pem</ssl_manager_cert>
    <ssl_manager_key>etc/certs/remoted-key.pem</ssl_manager_key>
  </auth>
```

has no `<force>` element, and the effective configuration read through the framework confirms nothing supplies one by default:

```
{ "auth": { "disabled": "no", "port": "1515", "use_source_ip": "no", "purge": "yes",
            "use_password": "yes", "ciphers": "...", "ssl_verify_host": "no",
            "ssl_manager_cert": "etc/certs/remoted.pem",
            "ssl_manager_key": "etc/certs/remoted-key.pem" } }
```

`wazuh-manager-authd` carries the matching refusals:

```
   ERROR: Duplicate agent name: %s
   ... already exists in the database.
   ... cannot be removed because the force option is disabled.
   ... cannot be replaced since it is not disconnected.
   ... has not been disconnected long enough to be replaced.
```

So a recreated container with a pinned name has no key, cannot enroll, and stays silent. `restart: always` will not help: the failure is a refusal from the manager, not a crash.

`<use_source_ip>no</use_source_ip>` closes the remaining escape: even the IP-based replacement path is off, and a container's IP changes anyway.

## Suggested fix

The agent compose needs to persist the identity, which is one volume:

```yaml
    volumes:
      - wazuh_agent_etc:/var/ossec/etc
```

so `client.keys` and the enrollment configuration survive a recreation and no second enrollment is attempted. With that in place, pinning `WAZUH_AGENT_NAME` becomes safe and is worth doing as well, since it makes the agent's identity independent of the container id.

If a deliberately stateless agent is the intent, then the manager side has to allow the replacement, which means shipping a `<force>` block in the `<auth>` configuration:

```xml
    <force>
      <enabled>yes</enabled>
      <disconnected_time enabled="yes">1h</disconnected_time>
      <after_registration_time>0</after_registration_time>
    </force>
```

and documenting that container agents must pin their name. What cannot stand is the current combination, where the agent is stateless, its name is derived from a value that changes, and the manager keeps every identity it ever handed out.

## Environment

- Agent 5.0.0, image `wazuh/wazuh-agent:5.0.0-latest`; manager 5.0.0, image `wazuh/wazuh-manager:5.0.0-latest`.
- `wazuh-docker` 5.0.0 branch at `37452c5c` (2026-08-24): `wazuh-agent/docker-compose.yml` and `build-docker-images/wazuh-agent/config/0-wazuh-init`.
- Effective manager configuration read with `wazuh.core.configuration.get_ossec_conf(section='auth')` in the running manager.

fcaffieri:
### Related

- https://github.com/wazuh/wazuh-docker/issues/2632

### Tests

- https://github.com/wazuh/wazuh-docker/issues/2632#issuecomment-5608794168


jotamolas (dismissed):
# Review

Read the diff in full — this replicates the manager's own `permanent_data`/exclusion mechanism for the agent (`mount_permanent_data`/`apply_exclusion_data`, build-time snapshot via `permanent_data.sh`), keying the persisted volume off `/var/ossec/etc` and excluding only `internal_options.conf` so it keeps refreshing from the image on every recreation — same reasoning the manager already relies on. The `find ... | read` empty-check and the `cp -ar .../.` copy idiom are both correct.

Federico's evidence in #2632 is thorough: 4 agent containers against the same manager (2 fixed, 2 control), each recreated 2-3 times, checked against both `client.keys` and the manager's own agent list each time — including a targeted test that edits both a persisted file and the excluded one before a recreation, confirming one survives and the other correctly reverts.

One thing worth flagging before this goes further: **the PR is now `CONFLICTING`** against `5.0.0` — #2654 (the CA verification fix for #2619) merged in the meantime, and both touch `0-wazuh-init`. The only real conflict is trivial (both branches append a line right after `set_manager_conn()` in `main()` — `set_agent_ssl(...)` from #2654 and `rm -rf .../data_tmp` from #2653 — no semantic overlap, both can just coexist). Everything else in the diff merges clean.

Also worth a note for final review: nobody has run these two fixes together in the same container yet (CA material under `/var/ossec/etc/certs` now lives inside the persisted volume too, once merged) — should be harmless since `set_agent_ssl` rebuilds its `<ssl>` block idempotently on every boot regardless of what's persisted, but it hasn't been exercised end-to-end as a combination.

LGTM.

jotamolas (approved):
LGTM

Issue #2629, opened 2026-09-04
Every environment variable the dashboard image declares is silently ignored


Rebits:
## Summary

The dashboard image declares thirteen configuration environment variables through `ENV` in its Dockerfile, which is how a container image advertises the knobs it accepts. `docker inspect` lists all thirteen. None of them has any effect. `wazuh_dashboard_config.sh` writes a setting into `opensearch_dashboards.yml` only when the key is already present in that file, and none of the thirteen keys is in the file the image ships, so each value is dropped with no message and an exit code of 0.

## The mechanism

`/wazuh_dashboard_config.sh` builds a map of 30 settings from environment variables and then applies them:

```bash
for key in "${!CONFIG_MAP[@]}"; do
    value="${CONFIG_MAP[$key]}"
    if [ -z "$value" ]; then
        continue
    fi
    escaped_key=$(echo "$key" | sed 's/[.[\*^$()+?{|]/\\&/g')

    # Try to replace existing line (commented or uncommented)
    if grep -q "^[#[:space:]]*${escaped_key}:" "$DASHBOARD_CONFIG_FILE"; then
        sed -i "s|^[#[:space:]]*${escaped_key}:.*|${key}: ${value}|" "$DASHBOARD_CONFIG_FILE"
    fi
done
```

The `if grep -q` has no `else`. A key that is not already in the file, commented or not, is skipped in silence.

The shipped `opensearch_dashboards.yml` is 27 lines and carries 16 of the 30 keys, including two commented placeholders for the indexer credentials. Comparing the two lists:

```
=== present in opensearch_dashboards.yml (the variable funciona): 16
   server.host, server.port, opensearch.hosts, opensearch.ssl.verificationMode,
   opensearch.username, opensearch.password, opensearch.requestHeadersAllowlist,
   opensearch_security.multitenancy.enabled, opensearch_security.readonly_mode.roles,
   server.ssl.enabled, server.ssl.key, server.ssl.certificate,
   opensearch.ssl.certificateAuthorities, uiSettings.overrides.defaultRoute,
   opensearch_security.session.ttl, opensearch_security.session.keepalive

=== ABSENT (the variable is silently ignored): 14
   opensearch_security.cookie.ttl                 <- OPENSEARCH_SECURITY_COOKIE_TTL
   pattern                                        <- PATTERN
   checks.pattern                                 <- CHECKS_PATTERN
   checks.template                                <- CHECKS_TEMPLATE
[...]

PR #2655, opened 2026-09-10
Fixed manager/dashboard environment variable attribution and documented missing manager variables

### LINKED ISSUE #2638: The deployment documentation attributes five environment variables to the wrong container and describes an override rule that works for neither file
## Summary

`docs/ref/configuration/environment-variables.md` is the only place the container deployment documents its configuration surface, and three of its four component sections are wrong. The manager section lists three variables the manager never reads, the dashboard section lists two it never reads, and seven variables the manager does read are documented nowhere, including `WAZUH_CLUSTER_KEY`, the one that would replace the cluster key baked into the public image. The final section then describes a general rule for overriding any configuration key through an environment variable; it does not hold for the dashboard, where only sixteen fixed keys are handled, nor for the indexer, whose entrypoint rejects the uppercase form the document prescribes and accepts only the literal lowercase key.

## The variables are attributed to the wrong containers

The document gives the manager and the dashboard the same five-line YAML block. Checking each name against the scripts each image actually runs, `0-wazuh-init`, `1-manager` and `entrypoint.sh` for the manager, `entrypoint.sh` and `wazuh_dashboard_config.sh` for the dashboard:

```
=== variables that the documentacion atribuye al MANAGER
   INDEXER_USERNAME     the lee
   INDEXER_PASSWORD     the lee
   WAZUH_API_URL        NO appears in its scripts
   DASHBOARD_USERNAME   NO appears in its scripts
   DASHBOARD_PASSWORD   NO appears in its scripts

=== variables that the documentacion atribuye al DASHBOARD
   INDEXER_USERNAME     NO appears in its scripts
   INDEXER_PASSWORD     NO appears in its scripts
   WAZUH_API_URL        the lee
   DASHBOARD_USERNAME   the lee
   DASHBOARD_PASSWORD   the lee
   API_USERNAME         the lee
   API_PASSWORD         the lee
```

The two blocks have been swapped in part: `WAZUH_API_URL`, `DASHBOARD_USERNAME` and `DASHBOARD_PASSWORD` belong to the dashboard and are documented under the manager, while `INDEXER_USERNAME` and `INDEXER_PASSWORD` belong to the manager and are documented under the dashboard. The descriptions follow the error, so the manager section explains `DASHBOARD_USERNAME / DASHBOARD_PASSWORD` as "Credentials for the Wazuh Dashboard to authenticate with the Indexer" while listing them as manager variables.

## Seven real manager variables are undocumented

Scanning the manager's own scripts for the operator-facing variables they expand:

```
   WAZUH_CLUSTER_BIND_ADDR
   WAZUH_CLUSTER_KEY
   WAZUH_CLUSTER_NODES
   WAZUH_CONFIG_MOUNT
   WAZUH_INDEXER_HOSTS
   WAZUH_NODE_NAME
   WAZUH_NODE_TYPE
```

None appears in the document, and two of them carry consequences that are documented nowhere else either:

- `WAZUH_CLUSTER_KEY` is the only way to replace `09184297c04d75e805538534681511a7`, the cluster key the public image ships and both compose files inherit. An operator reading the deployment documentation has no way to learn the variable exists. See [263](already-reported/BUG_263-the-cluster-key-that-authenticates-master-worker-communication-is-a-constant-in-the-public-image.md).
- `WAZUH_INDEXER_HOSTS` is used by both shipped compose files, and its value must be exactly `host:port`; every other plausible form is silently corrupted. See [261](already-reported/BUG_261-five-of-the-eight-plausible-values-for-wazuh-indexer-hosts-produce-a-corrupt-host-and-the-init-reports-success.md). With the variable undocumented, the required format is documented too.

The agent section, by contrast, is correct: `WAZUH_MANAGER_SERVER`, `WAZUH_REGISTRATION_SERVER` and `WAZUH_AGENT_NAME` are all read by the agent's init.

## The general override rule does not hold

The document closes with a rule presented as applying to configuration files generally:

> To override configuration values from files such as `opensearch.yml` and `opensearch_dashboards.yml` using environment variables:
> 1. Convert the configuration key to uppercase.
> 2. Replace any dots (`.`) in the key with underscores (`_`).
> 3. Assign the corresponding value.
>
> | `discovery.type: single-node`        | `DISCOVERY_TYPE=single-node`   |
> | `opensearch.hosts: https://url:9200` | `OPENSEARCH_HOSTS=https://url:9200` |
> | `server.port: 5601`                  | `SERVER_PORT=5601`             |
>
> This approach allows you to configure the services dynamically via Docker without modifying internal files.

**For `opensearch_dashboards.yml` it applies to sixteen keys and no others.** `wazuh_dashboard_config.sh` handles a fixed map of settings and writes each one only if its key is already present in the file (see [260](already-reported/BUG_260-every-environment-variable-the-dashboard-image-declares-is-silently-ignored.md)). Running the image with four settings named exactly as the rule prescribes:

```
=== keys present after apply the documented method:
  logging.verbose                    0
  server.maxPayloadBytes             0
  server.port                        1
  opensearch_security.cookie.ttl     0
=== value of server.port:
server.port: 5601
```

`server.port`, one of the document's own examples, works. `logging.verbose` and `server.maxPayloadBytes`, both ordinary OpenSearch Dashboards settings, are not written at all, and neither is `opensearch_security.cookie.ttl`, which is in the script's map but absent from the shipped template.

**For `opensearch.yml` the rule is the wrong shape entirely.** The indexer entrypoint converts environment variables into `-E` settings, but only when the name is already a lowercase dotted key:

```bash
    # OpenSearch settings need to have at least two dot separated lowercase
    # words, e.g. `cluster.name`, except for `processors` which we handle
    # specially
    if [[ "$envvar_key" =~ ^[a-z0-9_]+\.[a-z0-9_]+ || "$envvar_key" == "processors" ]]; then
        if [[ ! -z $envvar_value ]]; then
        opensearch_opt="-E${envvar_key}=${envvar_value}"
```

`DISCOVERY_TYPE` does not match that pattern: it has no dot and it is not lowercase, so it is discarded. The correct form is the literal key, `discovery.type=single-node`, which is exactly what the shipped compose uses for every indexer setting it passes:

```yaml
      - bootstrap.memory_lock=true
      - network.host=0.0.0.0
      - node.name=wazuh.indexer
      - cluster.initial_cluster_manager_nodes=wazuh.indexer
      - plugins.security.allow_default_init_securityindex=true
```

So the document's first example is contradicted by the deployment file that sits beside it.

## Suggested fix

Move the three dashboard variables out of the manager section and the two manager variables out of the dashboard section, and add the seven undocumented manager variables with their required formats, `WAZUH_INDEXER_HOSTS` as `host:port[,host:port...]` and `WAZUH_CLUSTER_KEY` with a note that leaving it unset keeps the packaged default.

Split the override section in two, since the two files work differently: for the indexer, the literal lowercase key with at least one dot; for the dashboard, the enumerated list of settings its script supports, or, better, fix the script so the rule becomes true (appending a key that is absent instead of skipping it).

## Environment

- `wazuh-docker` 5.0.0 branch at `37452c5c` (2026-08-24): `docs/ref/configuration/environment-variables.md`, both compose files, and the three images' entrypoint and init scripts.
- Images `wazuh/wazuh-manager:5.0.0-latest`, `wazuh/wazuh-dashboard:5.0.0-latest`, `wazuh/wazuh-indexer:5.0.0-latest`.
- Reproduced against the images alone; no running deployment needed.

jesusph04:
## Description 

1. **Swapped variables corrected**: `WAZUH_API_URL`, `DASHBOARD_USERNAME`,   `DASHBOARD_PASSWORD` moved from the manager section to the dashboard   section; `INDEXER_USERNAME`, `INDEXER_PASSWORD` moved from the
   dashboard section to the manager section. Verified against the   actual scripts each image runs (`1-manager` for the manager,   `wazuh_dashboard_config.sh`/`entrypoint.sh` for the dashboard).

2. **7 missing manager variables documented**: `WAZUH_CLUSTER_BIND_ADDR`,   `WAZUH_CLUSTER_KEY`, `WAZUH_CLUSTER_NODES`, `WAZUH_CONFIG_MOUNT`,   `WAZUH_INDEXER_HOSTS`, `WAZUH_NODE_NAME`, `WAZUH_NODE_TYPE` — all   confirmed present in `0-wazuh-init`. `WAZUH_CLUSTER_KEY` and   `WAZUH_INDEXER_HOSTS` get an extra note on their consequences   (packaged default key, strict `host:port` format), per the issue.

3. **Override rule (point 3) — not rewritten, flagged instead**: added   a note at the top of that section stating the rule is wrong for   both the indexer and the dashboard, linking #2629. Per the issue's   own suggestion: @mjbonacci has #2629 (dashboard env vars silently   ignored) assigned, and depending on how he fixes it, the "true"   dashboard rule may change — better to document it once than twice.

## Validation

All variable names and their scripts cross-checked directly against `0-wazuh-init`, `1-manager`, `entrypoint.sh` and
`wazuh_dashboard_config.sh` on `5.0.0` (not just taken from the issue report) before writing the doc changes.

### DIFF
--- a/docs/ref/configuration/environment-variables.md
+++ b/docs/ref/configuration/environment-variables.md
@@ -23,16 +23,25 @@ The Wazuh Manager container accepts the following environment variables, which c
 environment:
   - INDEXER_USERNAME=wazuh-manager
   - INDEXER_PASSWORD=wazuh-manager
-  - WAZUH_API_URL=https://wazuh.manager
-  - DASHBOARD_USERNAME=kibanaserver
-  - DASHBOARD_PASSWORD=kibanaserver
+  - WAZUH_NODE_NAME=wazuh.manager
+  - WAZUH_NODE_TYPE=master
+  - WAZUH_CLUSTER_KEY=
+  - WAZUH_CLUSTER_NODES=
+  - WAZUH_CLUSTER_BIND_ADDR=0.0.0.0
+  - WAZUH_INDEXER_HOSTS=wazuh.indexer:9200
+  - WAZUH_CONFIG_MOUNT=/wazuh-config-mount
 ```
 
 **Variable Descriptions:**
 
 - `INDEXER_USERNAME` / `INDEXER_PASSWORD`: Credentials for accessing the Wazuh Indexer with `wazuh-manager` user or a user with the same permissions. The value shown is the default the image ships; change it on the first start and write the new one here. See [Credentials](../credentials.md).
-- `WAZUH_API_URL`: URL of the Wazuh API, used by other services for communication.
-- `DASHBOARD_USERNAME` / `DASHBOARD_PASSWORD`: Credentials for the Wazuh Dashboard to authenticate with the Indexer.
+- `WAZUH_NODE_NAME`: This node's cluster name, written to `<cluster><node_name>`. Defaults to the container hostname.
+- `WAZUH_NODE_TYPE`: Either `master` or `worker`, written to `<cluster><node_type>`. Any other value (including unset) is treated as `master`.
+- `WAZUH_CLUSTER_KEY`: The cluster authentication key shared by every master/worker node, written to `<cluster><key>`. The public image ships a fixed default key; leaving this unset keeps that default, which is the same for every deployment using the published image and should be replaced in production. See [#263](https://github.com/wazuh/wazuh-docker/issues/263).
+- `WAZUH_CLUSTER_NODES`: Space-separated list of worker node addresses/hostnames, written to `<cluster><nodes>`. Only meaningful on the master node.
+- `WAZUH_CLUSTER_BIND_ADDR`: Address the cluster service binds to, written to `<cluster><bind_addr>`. Defaults to `0.0.0.0`.
+- `WAZUH_INDEXER_HOSTS`: Comma-separated list of indexer nodes as `host:port` (for example `wazuh.indexer:9200`, or `host1:9200,host2:9200` for multiple nodes), written to `<indexer><hosts>`. Each entry must be exactly `host:port`: no scheme, no trailing slash, no other separator — any other shape either fails validation or produces a corrupt host. See [#2630](https://github.com/wazuh/wazuh-docker/issues/2630).
+- `WAZUH_CONFIG_MOUNT`: Path, inside the container, where a directory mounted with the manager's own configuration files is looked for and copied over the packaged ones on start. Defaults to `/wazuh-config-mount`.
 - `WAZUH_REMOTE_BIND_ADDR`: Address `remoted` listens on for agent traffic, written to `<remote><https><bind_addr>` and `<remote><legacy><local_ip>`. Defaults to `0.0.0.0`, since the packaged `127.0.0.1` would make the published `1517` and `1514` unreachable from outside the container.
 
 ---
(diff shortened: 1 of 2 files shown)


docs/ref/configuration/environment-variables.md
```diff
@@ -23,16 +23,25 @@ The Wazuh Manager container accepts the following environment variables, which c
...
-  - WAZUH_API_URL=https://wazuh.manager
-  - DASHBOARD_USERNAME=kibanaserver
-  - DASHBOARD_PASSWORD=kibanaserver
+  - WAZUH_NODE_NAME=wazuh.manager
+  - WAZUH_NODE_TYPE=master
+  - WAZUH_CLUSTER_KEY=
+  - WAZUH_CLUSTER_NODES=
+  - WAZUH_CLUSTER_BIND_ADDR=0.0.0.0
+  - WAZUH_INDEXER_HOSTS=wazuh.indexer:9200
+  - WAZUH_CONFIG_MOUNT=/wazuh-config-mount
```
vcerenu: This variable shouldn't be modifiable, so I don't think this example should be included—though an explanation of its usage should be.

docs/ref/configuration/environment-variables.md
```diff
@@ -218,6 +218,14 @@ happens: the agent enrolls through the manager connection it already has, so
 ---
 
 ## Overriding Configuration Files with Environment Variables
+> **Note:** The rule below does not hold as written for either file it names.
+> For the indexer, the literal lowercase dotted key is required instead (for
+> example `discovery.type=single-node`, not `DISCOVERY_TYPE=single-node`) — see
+> the indexer's own compose entries for working examples. For the dashboard,
+> only a fixed set of keys handled by `wazuh_dashboard_config.sh` are read; any
+> other key is silently ignored. This section is being revised — see
+> [#2629](https://github.com/wazuh/wazuh-docker/issues/2629).
```
vcerenu: This note should take into account that the errors in the mentioned issue need to be corrected, so it is not suitable for the final documentation.

jotamolas (approved):
LGTM

PR #2659, opened 2026-09-10
Add wazuh manager worker password check

### LINKED ISSUE #2622: Document how to change default Indexer/Manager passwords in Docker deployments
## Description

Part of the installation security hardening effort: [wazuh-indexer-plugins#1519](https://github.com/wazuh/wazuh-indexer-plugins/issues/1519) found that a stock 5.0 install ships several Wazuh Indexer users with password equal to username (`admin`, `wazuh-demo`, `wazuh-admin`, `wazuh-readonly`, etc.), some of them highly privileged. Every central component and installation method must document how to change its default credentials right after installation.

For Docker deployments this means documenting how to change:

* Wazuh Indexer's reserved internal users (see wazuh-indexer-plugins#1519 for the full list).
* Wazuh Manager's `wazuh` and `wazuh-wui` users ([wazuh#38862](https://github.com/wazuh/wazuh/issues/38862), the equivalent Server-side ask).
* The `INDEXER_USERNAME`/`INDEXER_PASSWORD` pair the manager container uses to reach the indexer (`single-node/docker-compose.yml:36`, currently `wazuh-manager`/`wazuh-manager`) and `DASHBOARD_PASSWORD` (`:103`), which need to stay in sync after a rotation.

Related to wazuh#38862 (Manager, same ask). Attached as a sub-issue of wazuh-indexer-plugins#1519.

> [!CAUTION]
> **Draft — open question before this can be fully scoped.** `wazuh-passwords-tool.sh` (the tool being restored with `--change-all` in `wazuh-installation-assistant`) assumes a bare-metal/VM package layout (e.g. `/var/wazuh-manager/bin/wazuh-manager-keystore`). It is not yet confirmed whether it runs unmodified inside these Docker images, or whether the documented procedure here needs to be container-native instead — e.g. `docker exec` into the indexer container and re-run `securityadmin.sh` after editing `internal_users.yml` (the image already does this at bootstrap, see `build-docker-images/wazuh-indexer/config/securityadmin.sh`), or use the OpenSearch Security REST API (`PATCH _plugins/_security/api/internalusers/<user>`) instead. Confirm which approach applies before writing the final docs.

## Expected outcome

Immediately after `docker compose up`, an operator should have a clear, documented procedure to:

1. Change the password of every reserved Wazuh Indexer user.
2. Change the password of the Manager's `wazuh` and `wazuh-wui` users.
3. Update `INDEXER_USERNAME`/`INDEXER_PASSWORD` (and `DASHBOARD_PASSWORD`) to match, so containers keep talking to each other after the rotation.
4. Do all of this without breaking the manager-to-indexer connection or the dashboard.

## Tasks

- [ ] Confirm whether `wazuh-passwords-tool.sh --change-all` (once restored) works unmodified inside the Docker images, or document a container-native alternative.
- [ ] Document the procedure to change Wazuh Indexer default passwords.
- [ ] Document the procedure to change the Manager's `wazuh` and `wazuh-wui` passwords.
- [ ] Document how to update `INDEXER_USERNAME`/`INDEXER_PASSWORD`/`DASHBOARD_PASSWORD` after rotating, so the stack keeps working.
- [ ] Verify the documented procedure against a clean `docker compose up` (single-node and multi-node).

## Definition of Done

- [ ] Password-change instructions are documented immediately after the installation instructions.
- [ ] The procedure is verified end-to-end on a clean install: passwords change, manager/dashboard/indexer keep working.

## Reference

- wazuh-indexer-plugins#1519 (main security report)
- wazuh#38862 (Manager, same ask)
- Slack thread, Álex Ruiz, 2026-09-03

vcerenu:
Related issue https://github.com/wazuh/wazuh-docker/issues/2622

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 | Issue | Comment |
 | - | - |
+| [#2622](https://github.com/wazuh/wazuh-docker/issues/2622) | Check the Wazuh API accounts of every manager node in `check-default-credentials.sh`, and document the worker step of the multi-node password change |
 | [#6058](https://github.com/wazuh/internal-devel-requests/issues/6058) | Remove the OpenSearch demo users from the indexer image, add `password-tool.sh` to change the indexer and API passwords of a running deployment, and stop publishing the indexer port `9200` |
 | [#2601](https://github.com/wazuh/wazuh-docker/issues/2601) | Regenerate the manager self-signed server certificate per container at first boot |
 | [#2564](https://github.com/wazuh/wazuh-docker/issues/2564) | Add default AI assistant encryption key in the post installation script |
--- a/docs/ref/credentials.md
+++ b/docs/ref/credentials.md
@@ -234,10 +234,17 @@ printf '%s\n' '<the wazuh-wui password it printed>' | \
   docker compose exec -T wazuh.worker /password-tool.sh --user wazuh-wui --stdin
 ```
 
+**Do not skip the worker.** The API daemon runs on the master only, so a worker
+serves no API and nothing about the deployment looks wrong while its accounts
+are still on the defaults. What that database is for is the moment the worker
+becomes the master: a promoted node with an untouched database starts answering
+with `wazuh` and `wazuh-wui` as their own passwords. The step is what keeps a
+failover from undoing this procedure.
+
 ## Checking that no default is left
 
 ```bash
-cd single-node
+cd single-node        # or multi-node
 ../tools/tests/check-default-credentials.sh
 ```
 
@@ -245,6 +252,13 @@ It asserts that the image carries none of the OpenSearch demo accounts and that
 no Wazuh indexer or Wazuh API account authenticates with its own username as its
 password. **A deployment that has not been through the procedure above fails
 this check**, which is what it is for.
+
+The Wazuh API accounts are checked in two places: on the published API, and in
+the user database of every manager node the Compose file defines. The second one
+is what covers a worker, whose accounts no API answers for. A worker that has
+never been through the step above has no user database at all, and the check
+reports that as well, because the database a promotion creates holds the
+defaults.
 
 ## Notes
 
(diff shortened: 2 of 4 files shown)


jotamolas (approved):
# Review

Reviewed the diff in full — well-reasoned change:

- `service_images()` replaces a brittle sed/awk parse of `docker compose config` with something more robust, reused for both indexer image lookup and manager service discovery.
- `MANAGER_SERVICES` correctly picks up every service running a `wazuh-manager` image (confirmed `wazuh.master` and `wazuh.worker` both use `wazuh/wazuh-manager:5.0.0` in `multi-node/docker-compose.yml`).
- `api_db_state()` reads `rbac.db` directly via sqlite in **read-only** mode (so the check itself never creates the file), using the same `werkzeug.security.check_password_hash` the framework uses — not a reimplementation.
- Also fixes a real pre-existing bug: previously any non-200 HTTP code (including a transient `429`) was counted as "refused" (pass). Now only `401` passes, with retries for `429` specifically, and anything else fails with a clear reason.

## Verification

Built a patched `wazuh/wazuh-manager:5.0.0` locally (`FROM` the existing image, layering in the real `password-tool.sh` + `wazuh-api-users.py` from this branch — the published tags available to me don't have `#6058`'s changes yet) and stood up a real `multi-node` deployment:

1. Ran `docker compose exec wazuh.master /password-tool.sh --all` for real — rotated `wazuh`/`wazuh-wui`.
2. Ran the exact two `--stdin` commands the tool itself prints, against `wazuh.worker`, to set the same passwords there.
3. Ran `../tools/tests/check-default-credentials.sh`:

```
Wazuh API accounts of each manager node
  ok    wazuh.master: wazuh does not have 'wazuh' as its password
  ok    wazuh.master: wazuh-wui does not have 'wazuh-wui' as its password
  ok    wazuh.worker: wazuh does not have 'wazuh' as its password
  ok    wazuh.worker: wazuh-wui does not have 'wazuh-wui' as its password
```

Also confirmed the negative case before rotating the worker: `wazuh.worker: no Wazuh API user database, so wazuh would be seeded with 'wazuh' as its password` — the exact failure mode this PR exists to catch.

LGTM.

PR #2688, opened 2026-09-18
Add scheme test on wazuh manager init

### LINKED ISSUE #2635: The manager cannot be pointed at a plain HTTP indexer, and the TLS it insists on needs three certificates the image never creates
## Summary

`WAZUH_INDEXER_HOSTS` is the documented way to tell the containerised manager where its indexer is. The init script takes the value, throws away any scheme and writes `https://` unconditionally:

```
   /etc/cont-init.d/0-wazuh-init, lines 266 to 282

   if [[ -n "$WAZUH_INDEXER_HOSTS" ]]; then
       ...
       IFS=',' read -ra NODES <<< "$WAZUH_INDEXER_HOSTS"
       for NODE in "${NODES[@]}"; do
           IP="${NODE%:*}"
           PORT="${NODE#*:}"
           echo "      <host>https://$IP:$PORT</host>"        <- scheme hard coded
       done
```

There is no variable that selects the scheme, and because the host and port are split with
`${NODE%:*}` and `${NODE#*:}`, any value that contains one is mangled. Measured, one container
per row:

```
   WAZUH_INDEXER_HOSTS                what the init writes
   wazuh.indexer:9200                 <host>https://wazuh.indexer:9200</host>
   http://wazuh.indexer:9200          <host>https://http://wazuh.indexer://wazuh.indexer:9200</host>
   https://wazuh.indexer:9200         <host>https://https://wazuh.indexer://wazuh.indexer:9200</host>
```

So the bare form is silently forced to TLS, and the two URL forms, which are what the variable's
name invites, produce a host that is not an address at all. Passing the scheme the deployment
actually uses makes things worse than omitting it.

Having forced TLS, the shipped `<ssl>` block then names three files the image does not contain:

```
   <indexer>
     <hosts><host>https://wazuh.indexer:9200</host></hosts>
     <ssl>
       <certificate_authorities><ca>etc/certs/root-ca.pem</ca></certificate_authorities>
       <certificate>etc/certs/indexer-connector.pem</certificate>
       <key>etc/certs/indexer-connector-key.pem</key>
     </ssl>
   </indexer>
```

and the manager stops on start:

```
   wazuh-manager-analysisd: ERROR: An error occurred while initializing the modules:
   Could not initialize Indexer Connector: The CA root certificate file:
   'etc/certs/root-ca.pem' does not exist..
```

## The certificate directory after a normal start

The entrypoint generates two pairs and neither is the one the indexer connector is configured to use:

```
   apid-key.pem       apid.pem          generated
   remoted-key.pem    remoted.pem       generated
   root-ca.pem                          named in the configuration, absent
   indexer-connector.pem                named in the configuration, absent
   indexer-connector-key.pem            named in the configuration, absent
```

So on the published image the indexer connector cannot start under any environment: without certificates TLS fails, and there is no way to ask for plain HTTP instead.

## What it costs

`analysisd` is the process that decodes events and writes them to the indexer. With the connector unable to initialise, the manager runs a reduced set of daemons and nothing reaches the indexer: no events, no findings, no state, and no content namespace unpacked. A deployment in this state looks alive, because the container is up and the API answers, and produces nothing.

Measured on this stack before the workaround, after ten minutes: `wazuh-events-v5-*` 0 documents, `wazuh-findings-v5-*` 0, `wazuh-states-*` 0, and no `cmsync_standard_*` directory. After rewriting the block to plain HTTP, the same stack reached 508 decoders, 328 rules and 2,097 state documents.

## Reproducing

Point the variable at an http endpoint and read back what was written:

```
for v in wazuh.indexer:9200 http://wazuh.indexer:9200 https://wazuh.indexer:9200; do
  echo -n "$v -> "
  docker run --rm -e WAZUH_INDEXER_HOSTS="$v" -e WAZUH_CLUSTER_BIND_ADDR=0.0.0.0 \
    --entrypoint /bin/bash public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest \
    -c 'bash /etc/cont-init.d/0-wazuh-init >/dev/null 2>&1;
        sed -n "/<indexer>/,/<\/indexer>/p" /var/wazuh-manager/etc/wazuh-manager.conf |
        grep -o "<host>.*</host>"'
done
```

reproduces the three rows above.

The certificates the configuration names, against the ones that exist:

```
docker exec <manager> ls -la /var/wazuh-manager/etc/certs/
docker exec <manager> sed -n '/<indexer>/,/<\/indexer>/p' \
  /var/wazuh-manager/etc/wazuh-manager.conf
```

and the failure itself:

```
docker exec <manager> grep -i 'Indexer Connector' /var/wazuh-manager/logs/wazuh-manager.log
```

## Suggested fix

Honour a scheme when one is given, and default to `https` only when none is:

```
   case "$NODE" in
     http://*|https://*) echo "      <host>$NODE</host>" ;;
     *)                  echo "      <host>https://$NODE</host>" ;;
   esac
```

and drop the `<ssl>` block when the resolved scheme is `http`, or accept a
`WAZUH_INDEXER_SSL_VERIFICATION_MODE=none` in the way the dashboard image already accepts
`OPENSEARCH_SSL_VERIFICATION_MODE`.

Independently of the scheme, the image should either generate the three certificates its own
configuration names or stop naming them, because as shipped the connector cannot start even
against a TLS indexer.

## Environment

- `public.ecr.aws/wazuh-cicd/wazuh/wazuh-manager:5.0.0-latest`, pulled 2026-09-04, built 2026-09-04T03:52:58Z, `stage: rc1`, commit 7990fc0.
- Indexer: `wazuh-indexer:5.0.0-latest` with `plugins.security.disabled: true`, serving plain HTTP on 9200, which is the configuration this environment has used throughout.
- Worked around here with an entrypoint wrapper that rewrites the `<indexer>` block after the image's own init has run, in `/home/rebits/Wazuh/docker-stack/manager-entrypoint.sh`.

## Related

- [313](BUG_313-the-published-manager-image-never-starts-its-own-init-writes-an-empty-cluster-bind-address-that-its-schema-rejects.md) stops the same image before it reaches this point.
- [315](BUG_315-no-agent-can-enrol-against-the-published-manager-because-it-signs-its-own-enrolment-certificate-and-the-agent-demands-full-verification.md) is the third blocker in the same deployment.
- [281](already-reported/BUG_281-the-shipped-images-turn-off-tls-hostname-verification-in-three-places-although-the-certificates-the-product-generates-carry-the-right-names.md) is the same product's TLS defaults seen from the dashboard side.

vcerenu:
Related issue https://github.com/wazuh/wazuh-docker/issues/2635


CarlosALgit (dismissed):

jotamolas (approved):
LGTM

PR #2689, opened 2026-09-18
Add WAZUH_ENROLLMENT_TOKEN support to the agent

### LINKED ISSUE #2675: docker-agent package generation failed during "Build Wazuh images" step
During [2026-09-16 Nightly](https://github.com/wazuh/devel-major-incidents/issues/252), packages generation failed for docker-agent:

```
 > [linux/amd64 builder 4/6] RUN conf=/var/ossec/etc/ossec.conf;     grep -q 'CHANGE_MANAGER_ENDPOINT' "$conf" ||         { echo "ERROR: CHANGE_MANAGER_ENDPOINT is missing from the packaged ossec.conf"; exit 1; };     grep -q 'CHANGE_AGENT_NAME' "$conf" ||         { echo "ERROR: CHANGE_AGENT_NAME is missing from the packaged ossec.conf"; exit 1; }:
0.144 ERROR: CHANGE_MANAGER_ENDPOINT is missing from the packaged ossec.conf
------
Dockerfile:59
--------------------
  58 |     # since <endpoint> is the only tag it configures.
  59 | >>> RUN conf=/var/ossec/etc/ossec.conf; \
  60 | >>>     grep -q 'CHANGE_MANAGER_ENDPOINT' "$conf" || \
  61 | >>>         { echo "ERROR: CHANGE_MANAGER_ENDPOINT is missing from the packaged ossec.conf"; exit 1; }; \
  62 | >>>     grep -q 'CHANGE_AGENT_NAME' "$conf" || \
  63 | >>>         { echo "ERROR: CHANGE_AGENT_NAME is missing from the packaged ossec.conf"; exit 1; }
  64 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c conf=/var/ossec/etc/ossec.conf;     grep -q 'CHANGE_MANAGER_ENDPOINT' \"$conf\" ||         { echo \"ERROR: CHANGE_MANAGER_ENDPOINT is missing from the packaged ossec.conf\"; exit 1; };     grep -q 'CHANGE_AGENT_NAME' \"$conf\" ||         { echo \"ERROR: CHANGE_AGENT_NAME is missing from the packaged ossec.conf\"; exit 1; }" did not complete successfully: exit code: 1
```

Action https://github.com/wazuh/wazuh-docker/actions/runs/35052064499/job/104654862220

### LINKED ISSUE #2676: Replace manual authd.pass/<endpoint> writing in 0-wazuh-init with WAZUH_ENROLLMENT_TOKEN support
## Description
wazuh/wazuh#39063 (merged to `5.0.0` via wazuh/wazuh#39285, 2026-09-15) replaces the classic `WAZUH_MANAGER`/`WAZUH_REGISTRATION_PASSWORD`/`WAZUH_REGISTRATION_CA` contract with a single `WAZUH_ENROLLMENT_TOKEN`, but unlike Ansible the agent container never runs the packaged installer — the package is already installed in the image, and `build-docker-images/wazuh-agent/config/etc/cont-init.d/0-wazuh-init` reimplements the enrollment bootstrap by hand: it writes the `<manager><endpoint>` block via `sed` in `resolve_manager_endpoint` (starting around line 192), and it writes `authd.pass` itself in `set_agent_name_and_password()` (defined at line 303), which checks `WAZUH_REGISTRATION_PASSWORD` at line 307 and writes the file at lines 314-315. None of this goes through `register_configure_agent.sh`, so nothing here gets deprecated automatically — it keeps working exactly as before against a manager that no longer has a static password to accept, and enrollment fails.

## Root cause
`0-wazuh-init` predates the token model and never delegates to the shared installer script, so it needs its own real implementation of the token bootstrap instead of inheriting the new behavior for free the way a script-driven install would.

## Suggested fix
- Accept a `WAZUH_ENROLLMENT_TOKEN` environment variable in `0-wazuh-init`.
- When it's present, write it verbatim to `/var/ossec/etc/enrollment_token` with mode 600, owner `root:root`, and stop writing `<endpoint>` by hand in `resolve_manager_endpoint`, since the manager address is already encoded in the token.
- Decide whether to keep the current `WAZUH_REGISTRATION_PASSWORD` path as an explicit 4.x fallback or drop it outright — this needs a call from the team, not assumed here.
- Same open question shared with the other two issues in this batch: who mints the token and when — here it's harder still, since docker-compose has no "manager mints, agent consumes" handoff step set up today at all.

## Validations
- [x] Agent container enrolls successfully using only `WAZUH_ENROLLMENT_TOKEN`, with no `WAZUH_REGISTRATION_PASSWORD` set.
- [x] `/var/ossec/etc/enrollment_token` is written with mode 600 and owner `root:root`.
- [x] The `<endpoint>` tag is no longer written manually when a token is used.
- [x] Whatever fallback decision is made for 4.x is verified against an actual 4.x manager.

## Reference
- wazuh#39063 — https://github.com/wazuh/wazuh/issues/39063
- wazuh#39285 — https://github.com/wazuh/wazuh/pull/39285
- `docs/ref/modules/authd/README.md` and `docs/ref/modules/authd/enrollment-lifecycle.md` (branch `5.0.0`, `wazuh/wazuh`)
- `build-docker-images/wazuh-agent/config/etc/cont-init.d/0-wazuh-init:303-321`
- `wazuh-qa-automation#8999` — blocked on the same missing guide as this issue (wazuh#39068)

jotamolas:
## Description

wazuh/wazuh#39063 makes the enrollment token the only way to register a 5.x agent, and removes the address/credential variables `register_configure_agent.sh` used to accept. The agent container never runs that installer, so `0-wazuh-init` needed its own implementation.

Related #2676. Also fixes #2675 (docker-agent build failure), same root cause.

## Changes

- `0-wazuh-init`: decode `WAZUH_ENROLLMENT_TOKEN` via `wazuh-agentd --show-token`, write `<endpoint>` from it, store the token at `etc/enrollment_token` (`0600 root:root`) for the agent's own bootstrap. Drop `authd.pass`/`WAZUH_REGISTRATION_PASSWORD` — no 4.x fallback, a 5.x agent only ever talks to a 5.x manager.
- `Dockerfile`: write the `CHANGE_MANAGER_ENDPOINT` placeholder ourselves instead of relying on the installer, which no longer does it without a token.
- Docs and the `wazuh-agent` compose example updated to lead with the token.

## Test plan

- [x] Built the image and confirmed the placeholder fix in the Dockerfile against the real package.
- [x] Ran the agent against a real 5.0.0 manager with a token minted via `POST /agents/enrollment-tokens`: enrolls, shows up `active` on the manager, `enrollment_token` deleted after bootstrap.
- [x] Restarted the same enrolled container: no re-enrollment, no duplicate agent.
- [x] Token + manual endpoint together, no manager configured at all, and a rejected token all fail with a clear message.
- [x] Manual `WAZUH_MANAGER_ENDPOINT`/`WAZUH_MANAGER_SERVER` path still works unchanged.


vcerenu (approved):
LGTM

